PR #56, opened 2017-07-29
Reference rewrite

stohrendorf:


TombEditor/Clipboard.cs
```diff
@@ -39,80 +35,80 @@ public static void Copy()
...
-            if (_editor.PickingResult.ElementType == PickingElementType.Block)
+            if (_editor.PickingResult._elementType == PickingElementType.Block)
             {
-                int x = _editor.PickingResult.Element >> 5;
-                int z = _editor.PickingResult.Element & 31;
+                int x = _editor.PickingResult._element >> 5;
+                int z = _editor.PickingResult._element & 31;
 
-                Block block = _editor.Level.Rooms[room].Blocks[x, z];
+                Block block = room.Blocks[x, z];
```
TeslaRus: can be room invalid here? (no selected room)

TombEditor/Geometry/Block.cs
```diff
@@ -58,49 +62,43 @@ public bool IsFloorSplit
...
-                int x = 0;
-                int z = 0;
+                const int x = 0;
+                const int z = 0;
 
-                Vector3 p1 = new Vector3(x * 1024.0f, QAFaces[0] * 256.0f, z * 1024.0f);
-                Vector3 p2 = new Vector3((x + 1) * 1024.0f, QAFaces[1] * 256.0f, z * 1024.0f);
-                Vector3 p3 = new Vector3((x + 1) * 1024.0f, QAFaces[2] * 256.0f, (z + 1) * 1024.0f);
-                Vector3 p4 = new Vector3(x * 1024.0f, QAFaces[3] * 256.0f, (z + 1) * 1024.0f);
+                var p1 = new Vector3(x * 1024.0f, QAFaces[0] * 256.0f, z * 1024.0f);
```
TeslaRus, line 68: may be add names for the magic like 1024 and 256...

PR #71, opened 2017-08-02
Portal reference

stohrendorf:


TombEditor/Compilers/Structs.cs
```diff
@@ -545,7 +545,7 @@ public struct tr_overlap_aux
         public bool IsEdge;
     }
 
-    public struct tr_sector_aux
+    public class TrSectorAux
```
ActuallyaDeviloper, line 549: Is there a specific reason why you changed this class to a struct?
It seems none beneficial to me.

TombEditor/Geometry/IO/Prj2Writer.cs
```diff
@@ -72,12 +73,12 @@ public static bool SaveToPrj2(string filename, Level level)
...
                     // Write portals
-                    int numPortals = level.Portals.Count;
+                    int numPortals = level.Portals.Count();
                     writer.Write(numPortals);
-                    foreach (var p in level.Portals.Values)
+                    foreach (var p in level.Portals)
                     {
                         writer.Write(p.Id);
-                        writer.Write(p.OtherId);
+                        writer.Write((int) p.Other.Id);
```
ActuallyaDeviloper: I think you started introducing those in the first place: Can you please refrain from putting spaces after casts? This looks unconventional to me and violates the existing coding style. Additionally the standard setting of Visual Studio is the other way, usually code on the internet (eg code samples on [MSDN](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/types/casting-and-type-conversions)) don't have them either.

TombEditor/Geometry/Room.cs
```diff
@@ -27,7 +28,11 @@ public class Room
         public List<int> Sinks { get; private set; } = new List<int>();
         public List<int> Cameras { get; private set; } = new List<int>();
         public List<int> FlyByCameras { get; private set; } = new List<int>();
-        public List<int> Portals { get; private set; } = new List<int>();
+        public IEnumerable<Portal> Portals => Enumerable
+            .Range(0, Blocks.GetLength(0))
+            .SelectMany(x => Enumerable.Range(0, Blocks.GetLength(1))
+                .SelectMany(y => new []{Blocks[x,y].CeilingPortal, Blocks[x,y].FloorPortal, Blocks[x,y].WallPortal})
+        ).Where(p => p != null);
```
ActuallyaDeviloper: Avoiding any state here altogether, looks extremely beautiful to me.

However I am slightly worried about performance. LINQ is not famous for being fast and some code might still do some crazy indexing into this. Have you done any investigation about how often this method is actually called?

Also this looks to me like the same portals are enumerated multiple times. 
I believe .Distinct() should be added.

TombEditor/Compilers/Structs.cs
stohrendorf: Structs are copied by-value, which means that changing members within a function would not affect the outside instance, but this is indeed necessary now.

TombEditor/Geometry/Room.cs
stohrendorf: Haven't really looked into this, but it needs investigation, true. On the other hand, I haven't noticed any lags yet.

TombEditor/Geometry/ObjectInstance.cs
```diff
@@ -12,22 +13,43 @@ public enum ObjectInstanceType : byte
...
+            return Id == other.Id;
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
+            var other = obj as ObjectInstance;
+            return other != null && Equals(other);
+        }
```
ActuallyaDeviloper: Is there a specific reason why you added this code?
As far as I understand the object system different object instances should always have different IDs too.

TombEditor/Utils.cs
```diff
@@ -224,21 +226,28 @@ public static int ReferenceIndexOf<T>(this IList<T> list, T needle)
...
                     return i;
 
             return -1;
         }
 
+        public static IEnumerable<T> Unwrap<T>(this T[,] array)
+        {
+            for (int x = 0; x < array.GetLength(0); ++x)
+            for (int y = 0; y < array.GetLength(1); ++y)
+                yield return array[x, y];
```
ActuallyaDeviloper: Please indent the second loop.

TombEditor/Utils.cs
```diff
@@ -55,23 +54,25 @@ public static void ConvertTextureTo256Width(ref Bitmap bitmap)
                     int numYtiles = bitmap.Height / 64;
 
                     for (int yy = 0; yy < numYtiles; yy++)
-                        for (int xx = 0; xx < numXtiles; xx++)
+                    for (int xx = 0; xx < numXtiles; xx++)
+                    {
```
ActuallyaDeviloper: Please don't align nested control structures the same. (My reasoning here is the same as for the cast spaces, the spaces after casts)

TombEditor/Geometry/Room.cs
```diff
@@ -99,8 +100,8 @@ public void Resize(int numXSectors, int numZSectors, short ceiling = DefaultHeig
 
         public Vector2 SectorPos
         {
-            get { return new Vector2(Position.X, Position.Z); }
-            set { Position = new Vector3(value.X, Position.Y, value.Y); }
+            get => new Vector2(Position.X, Position.Z);
+            set => Position = new Vector3(value.X, Position.Y, value.Y);
         }
```
ActuallyaDeviloper: I have nothing against this aesthetically, however your use of => in this context produces an compiler error for me with Visual Studio 2015 (14.0.25422 Update 3)

PR #106, opened 2017-09-13
Implementation of a purely chunk based level format

ActuallyaDeviloper:
This is the start of such an implementation. Currently it is *only* implemented for the level settings which is not extremely exciting but it shows all required features. Also this basically defined every variable as its own 'Chunk' so the code and size overhead is maximal here. For most other code bits its propably much more realistic, to group code that belongs together.

General information about current implementation choices can be read in [Chunk.cs](https://github.com/MontyTRC89/Tomb-Editor/pull/106/files#diff-c3ccd81c73a7e173156e6cefb0e27f0f).
A partially complete suggestion for the file chunk tree can be found in [Prj2Chunks.cs](https://github.com/MontyTRC89/Tomb-Editor/pull/106/files#diff-68e6c774e2e76cb6d5c89ddac8b89d64). (Its visually structured like a tree, even thought the VS formatter very strongly rebels against that.)

An example of how lists of objects will be handled is given by the 'LevelTextures' which are now actually saved as an area. Currently there is still always only one, but its also a step in the direction of allowing more than one level texture.

While this this version is a prototype, I made sure that it actually works and it actually produces functioning *.prj2 files that can be loaded again.

### DIFF
--- a/TombEditor/Geometry/IO/Prj2Chunks.cs
+++ b/TombEditor/Geometry/IO/Prj2Chunks.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using TombLib.IO;
+
+namespace TombEditor.Geometry.IO
+{
+    // TODO Add documentation for binary offsets of the chunks
+    internal static class Prj2Chunks
+    {
+        public static readonly byte[] MagicNumber = new byte[] { 0x50, 0x52, 0x4A, 0x32 };
+        public const uint Version = 3;
+        public const uint VersionCompressed = Version ^ 0x80000000;
+
+        public static readonly ChunkID Settings = ChunkID.FromString("TeSettings");
+        /**/public static readonly ChunkID WadFilePath = ChunkID.FromString("TeWadFilePath");
+        /**/public static readonly ChunkID FontTextureFilePath = ChunkID.FromString("TeFontTextureFilePath");
+        /**/public static readonly ChunkID SkyTextureFilePath = ChunkID.FromString("TeSkyTextureFilePath");
+        /**/public static readonly ChunkID SoundPaths = ChunkID.FromString("TeSoundPaths");
+        /******/public static readonly ChunkID SoundPath = ChunkID.FromString("TeSoundPath"); //String
+        /**/public static readonly ChunkID GameDirectory = ChunkID.FromString("TeGameDirectory");
+        /**/public static readonly ChunkID GameLevelFilePath = ChunkID.FromString("TeGameLevelFilePath");
+        /**/public static readonly ChunkID GameExecutableFilePath = ChunkID.FromString("TeGameExecutableFilePath");
+        /**/public static readonly ChunkID GameExecutableSuppressAskingForOptions = ChunkID.FromString("TeGameExecutableSuppressAskingForOptions");
+        /**/public static readonly ChunkID IgnoreMissingSounds = ChunkID.FromString("TeIgnoreMissingSounds");
+        /**/public static readonly ChunkID Textures = ChunkID.FromString("TeTextures");
+        /******/public static readonly ChunkID Texture = ChunkID.FromString("TeTexture");
+        /**********/public static readonly ChunkID TexturePath = ChunkID.FromString("TePath");
+        /**********/public static readonly ChunkID TextureConvert512PixelsToDoubleRows = ChunkID.FromString("Te512C");
+        /**********/public static readonly ChunkID TextureReplaceMagentaWithTransparency = ChunkID.FromString("TeMagentaR");
+        public static readonly ChunkID Rooms = ChunkID.FromString("TeRooms");
+        /**/public static readonly ChunkID Room = ChunkID.FromString("TeRoom"); // Contains X, Y sectors, Name, Position directly
+        /******/public static readonly ChunkID Sectors = ChunkID.FromString("TeSectors");
+        /**********/public static readonly ChunkID Sector = ChunkID.FromString("TeS");
+        /**************/public static readonly ChunkID SectorGeometry = ChunkID.FromString("TeG"); //EDFaces, QAFaces, WSFaces, RFFaces, [*]SplitDirectionToggled, 
+        /**************/public static readonly ChunkID SectorProperties = ChunkID.FromString("TeP"); //Flags, Opacitiy, ...
+        /******/public static readonly ChunkID RoomProperties = ChunkID.FromString("TeRoomP"); // Ambient, Flags, ...
+        /******/public static readonly ChunkID Objects = ChunkID.FromString("TeObjects");
+        /**********/public static readonly ChunkID Camera = ChunkID.FromString("TeCam");
+        /**************/public static readonly ChunkID CameraData = ChunkID.FromString("TeDat");
+        /**************/public static readonly ChunkID Pos = ChunkID.FromString("TePos");
+        /**********/public static readonly ChunkID FlybyCamera = ChunkID.FromString("TeFlyBy");
+        /**************///public static readonly ChunkID Pos = ChunkID.FromString("TePos"); // Used here too
+        /**************/public static readonly ChunkID FlyCameraData = ChunkID.FromString("TeDat");
+        /**********/public static readonly ChunkID Light = ChunkID.FromString("TeLight");
+        /**************///public static readonly ChunkID Pos = ChunkID.FromString("TePos"); // Used here too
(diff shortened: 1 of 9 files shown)


TombEditor/Geometry/IO/Prj2Chunks.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace TombEditor.Geometry.IO
+{
+    // TODO Add documentation for binary offsets of the chunks
+    internal static class Prj2Chunks
+    {
+        public static readonly byte[] MagicNumber = new byte[] { 0x50, 0x52, 0x4A, 0x32 };
+        public const uint Version = 3;
+        public const uint VersionCompressed = Version ^ 0x80000000;
+
+        public static readonly ChunkID Settings = ChunkID.FromString("TeSettings");
```
stohrendorf, line 17: These IDs have a smurf naming convention. Is the "Te" prefix really necessary? Also, how do you ensure that IDs are unique?

TombLib/IO/Chunk.cs
```diff
@@ -0,0 +1,231 @@
...
+    //     - The minimal size overhead of introducing a new chunk is only 2 bytes + the byte length of the ChunkID
+    //
+
+
+    public struct ChunkID
+    {
+        public static readonly ChunkID Empty = new ChunkID(new byte[0], 0);
+        private static readonly Encoding _encoding = Encoding.UTF8;
+
+        private byte[] _idBytes; // Actually store the chunk ID as bytes. We don't want fancy Unicode comparison (different id bytes would compare identical, also big overhead)
```
stohrendorf, line 50: Simply require to use ASCII for chunk IDs.

TombLib/IO/LEB128.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
+
+namespace TombLib.IO
+{
+    // https://en.wikipedia.org/wiki/LEB128
+    internal static class LEB128
```
stohrendorf, line 6: This introduces complexity that shouldn't be necessary.

TombLib/IO/Chunk.cs
```diff
@@ -0,0 +1,231 @@
...
+            return " at offset " + chunkStart + " with size " + chunkSize + ". " + chunkID;
+        }
+
+        public delegate bool ParseChunkDelegate(BinaryReaderEx stream, ChunkID chunkID, long chunkSize);
+        public static void ParseChunks(BinaryReaderEx stream, ParseChunkDelegate tryParseChunk)
+        {
+            do
+            {
+                ChunkID chunkID = ChunkID.FromStream(stream);
+                if (chunkID == ChunkID.Empty) // End reached
```
stohrendorf, line 164: Why providing a null object, and not let `FromStream` return null in the first place? This logic requires users to always compare to the null object, and null itself is not a "special chunk" that can be handled like any other chunk.

TombLib/IO/Chunk.cs
```diff
@@ -0,0 +1,231 @@
...
+    //
+    // Key properties of the format:
+    //     - Unknown chunks can be ignored to enable backwards as well as forwards compability.
+    //     - Old chunk types can be replaced by new ones simply by choosing a new Id.
+    //     - An empty chunks stream is only 1 byte big. (A single 0 byte, so that the process is aborted immediately like described above)
+    //     - The minimal size overhead of introducing a new chunk is only 2 bytes + the byte length of the ChunkID
+    //
+
+
+    public struct ChunkID
```
stohrendorf, line 45: This seems a bit over-engineered.

TombLib/IO/Chunk.cs
```diff
@@ -0,0 +1,231 @@
...
+                    return "Invalid UTF-8 sequence";
+                }
+            }
+        }
+
+        public override string ToString()
+        {
+            StringBuilder str = new StringBuilder("Name: \"");
+            str.Append(Name);
+            str.Append("\", Hex: ");
```
stohrendorf, line 134: You don't need a hex representation if you simply require IDs to be in ASCII.

TombLib/IO/Chunk.cs
```diff
@@ -0,0 +1,231 @@
...
+    //         n bytes that identificate the chunk.
+    //
+    //         Custom ChunkIds can be defined, but they are recommended to ASCII or UTF-8 strings
+    //         to allow easy identification in a hex editor.
+    //
+    //         To avoid Id collisions (different people define the same chunk Ids to different things),
+    //         it is recommended to begin all custom ChunkIds of some software with a unique prefix.
+    //         (0x54, 0x65 ('Te') in case for Tomb Editor)
+    //                                   
+    //     LEB unsigned int (up to 9 bytes)  ChunkSize
```
stohrendorf: If you really enter the GB area you have a problem anyways, so just use a uint.

TombLib/IO/Chunk.cs
```diff
@@ -0,0 +1,231 @@
...
+    //         The size of the content of the chunk
+    //
+    //     n bytes                           ChunkContent
+    //         n bytes that identificate the chunk.
+    //         The meaning of those byte is entirely dependent on the previously read ChunkId.
+    //         Applications are encouraged to ignore chunk content whose IDs they don't recognize.
```
stohrendorf: sed s/encouraged/allowed/

TombEditor/Geometry/IO/Prj2Loader.cs
```diff
@@ -18,280 +18,319 @@ public static class Prj2Loader
...
+                    return false;
+                return true;
+            });
+            settings.Textures.Add(new LevelTexture(settings, path, convert512PixelsToDoubleRows, replaceMagentaWithTransparency));
+            return true;
+        }
 
+        public static bool LoadRooms(BinaryReaderEx reader, ChunkID idOuter, Level level)
+        {
+            if (idOuter != Prj2Chunks.Rooms)
```
stohrendorf, line 134: So every user/callee is required to check if they can process the chunk? That's not nice. Better provide a mapping from chunk IDs to the processors, or use a visitor, or refactor to check the IDs outside the callees.

TombEditor/Geometry/IO/Prj2Chunks.cs
ActuallyaDeviloper: Unfortunately we can't really *guarantee* that they are unique. The 'Te' prefix is just a very simple approach to distinguish between different vendors similar to how people prefix functions in C. We could include namespaces into the system but that seems like overengineering to me considering that would end up achieving the same thing as this.

An approach to achieve uniqueness to a certain degree is to use a random number for a prefix of sufficient length.

TombLib/IO/Chunk.cs
ActuallyaDeviloper: I deliberately have not designed this as a text format. Its just that the bytes we write currently happen to represent text encoded in ASCII, UTF-8 as well as codepage 1252. This is handy for debugging but not much beyond that.

In the future we or third party is free to use something else. (eg unique id bytes from a random number generator that I mentioned above)
We will then crash with values outside the valid range if we save those couple lines of code.

TombLib/IO/Chunk.cs
ActuallyaDeviloper: It is because there is actually a 0 written inside the file. You can check with an hex editor and there is a 0. There isn't just nothing, there is actually a 0. If it were a string, it would be "" instead of null.

TombLib/IO/LEB128.cs
ActuallyaDeviloper: You are right, it is not strictly necessary.

If file size wouldn't be a concern, I would scrap this.
With this the minimal overhead of a chunk is only 2 bytes.

I am also not particularly fixed on LEB128. We could also use another format with similar properties. It just seemed reasonable simple and since its already kind of a standard, people can read up on how to interpret it on the internet.

TombLib/IO/Chunk.cs
ActuallyaDeviloper: I changed the encoding to use ASCII instead of UTF-8 as default representation afterall.

TombEditor/Geometry/IO/Prj2Loader.cs
ActuallyaDeviloper: I admit, I don't like the current code pattern here currently either.

I am not entirely sure how the visitor pattern would apply here. Maybe you can provide an example?

Why am I not doing the checks outside the function?
It certainly would look nicer right now, but that would limit the reusability of those function in the long term. Over time the format will probably evolve and we will see some chunk types being replaced by others. By doing the check *inside* the function, the function can accept different chunk types without any further reaching changes.

PR #114, opened 2017-09-21
Dockable UI

Lwmte:
Introduces freeform UI with dockable tool windows. While original DarkUI library is pretty limited on docking, its features should be perfectly enough for our needs.

What it does:

1) All controls are grouped and decoupled into their own control windows.
2) There is default window layout which gets loaded on first editor launch or by menu entry in "Tools" menu.
3) Window layout is stored into configuration.
4) Main window maximized state, size and position are preserved into configuration.
5) [not actually dock-related] Added nifty texture loading helper into texture panel.

What it doesn't: you can't stack or manually resize control windows (DarkUI limitation). To get control size properly, you must define MinimumSize dimensions which correspond to actual control window size.

### DIFF
--- a/TombEditor/ToolWindows/TexturePanel.cs
+++ b/TombEditor/ToolWindows/TexturePanel.cs
@@ -21,8 +21,7 @@ namespace TombEditor.ToolWindows
 
             _editor = Editor.Instance;
             _editor.EditorEventRaised += EditorEventRaised;
-
-            panelTextureMap.Configuration = _editor.Configuration;
+            
             panelTextureMap.SelectedTextureChanged += delegate { _editor.SelectedTexture = panelTextureMap.SelectedTexture; };
         }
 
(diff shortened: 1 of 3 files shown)


TombEditor/ToolWindows/Lighting.cs
```diff
@@ -0,0 +1,215 @@
...
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+                _editor.EditorEventRaised -= EditorEventRaised;
+            if (disposing && (components != null))
+                components.Dispose();
+            base.Dispose(disposing);
+        }
+
+        public void Initialize()
```
ActuallyaDeviloper: Please move this code to the constructor.
The constructor is here to do initialization work.

TombEditor/ToolWindows/Palette.cs
```diff
@@ -0,0 +1,56 @@
...
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+                _editor.EditorEventRaised -= EditorEventRaised;
+            if (disposing && (components != null))
+                components.Dispose();
+            base.Dispose(disposing);
+        }
+
+        public void Initialize()
```
ActuallyaDeviloper: Please move this code to the constructor.

TombEditor/ToolWindows/TexturePanel.cs
```diff
@@ -0,0 +1,74 @@
...
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+                _editor.EditorEventRaised -= EditorEventRaised;
+            if (disposing && (components != null))
+                components.Dispose();
+            base.Dispose(disposing);
+        }
+
+        public void Initialize()
```
ActuallyaDeviloper: Please move this code to the constructor.

TombEditor/FormMain.cs
```diff
@@ -87,117 +77,45 @@ protected override void Dispose(bool disposing)
             base.Dispose(disposing);
         }
 
-        private void EditorEventRaised(IEditorEvent obj)
+        private void InitializeWindow()
```
ActuallyaDeviloper: This methode and the next one appear not as useful abstractions to me.
I suggest that we inline them.

PR #130, opened 2017-10-03
Beautify depth bar

Lwmte:
I tried to make depth bar more stylish, with semi-transparent selection and nifty arrows.
Also fixed a bug when depth bar was not redrawn correctly with window resize.

clipboard-1

### DIFF
--- a/TombEditor/Controls/Panel2DMap.cs
+++ b/TombEditor/Controls/Panel2DMap.cs
@@ -119,10 +119,10 @@ namespace TombEditor.Controls
             int currentProbeIndex = 0;
 
             for (int i = 0; i < _depthBar.DepthProbes.Count; ++i)
-                if ((clickPos - _depthBar.DepthProbes[i].Position).LengthSquared() < (clickPos - _depthBar.DepthProbes[currentProbeIndex].Position).LengthSquared())
+                if ((clickPos - _depthBar.DepthProbes[i].Position).Length() < (clickPos - _depthBar.DepthProbes[currentProbeIndex].Position).Length())
                     currentProbeIndex = i;
 
-            if ((clickPos - _depthBar.DepthProbes[currentProbeIndex].Position).LengthSquared() < _probeRadius / ViewScale)
+            if ((clickPos - _depthBar.DepthProbes[currentProbeIndex].Position).Length() < _probeRadius / 2 / ViewScale)
                 return currentProbeIndex;
             else
                 return null;


TombEditor/Controls/Panel2DMap.cs
```diff
@@ -146,14 +171,18 @@ protected override void OnMouseDown(MouseEventArgs e)
                 case MouseButtons.Middle:
                 case MouseButtons.XButton2:
                     _currentlyEditedDepthProbeIndex = _depthBar.DepthProbes.FindIndex((depthProbe) => (
-                       (Math.Round(depthProbe.X) == Math.Round(clickPos.X)) &&
-                       (Math.Round(depthProbe.Y) == Math.Round(clickPos.Y))));
+                       (depthProbe.Position.X - 2 < clickPos.X) &&
```
ActuallyaDeviloper: I think we should adopt the code from what we are doing on "XButton1".
Not just use any index, but the one closest to the mouse pointer.

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -111,6 +111,31 @@ private void MoveToFixedPoint(PointF visualPoint, Vector2 worldPoint)
             Invalidate();
         }
 
+        private void SetProbeColor(DepthBar.DepthProbe probe)
```
ActuallyaDeviloper: I don't like this abstraction. I think a GetProbeColor() would be better, the result can then also be used in the constructor directly.

Is there a reason why the color selection is randomized?
How about the following, that always picks the least used color instead:
```
        private Color GetProbeColor()
        {
            return _depthBar.DepthProbes
                 .Select(probe => probe.Color)
                 .Concat(DepthBar.ProbeColors) // Make a list of *all* colors
                 .GroupBy(color => color) // Group same colors
                 .OrderBy(group => group.Count()) // Sort color groups after how often they appeared
                 .First().First(); // Use first color of the least occurring color group.
        }
```

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -183,8 +212,8 @@ protected override void OnMouseDoubleClick(MouseEventArgs e)
                 case MouseButtons.XButton2:
                     Vector2 clickPos = FromVisualCoord(e.Location);
                     int currentProbeIndex = _depthBar.DepthProbes.FindIndex((depthProbe) => (
-                       (Math.Round(depthProbe.X) == Math.Round(clickPos.X)) &&
-                       (Math.Round(depthProbe.Y) == Math.Round(clickPos.Y))));
+                       (Math.Round(depthProbe.Position.X) == Math.Round(clickPos.X)) &&
+                       (Math.Round(depthProbe.Position.Y) == Math.Round(clickPos.Y))));
```
ActuallyaDeviloper: Same as above, even though this is not actually your code.
We should reuse the code from "XButton1".

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -361,13 +392,38 @@ protected override void OnPaint(PaintEventArgs e)
...
-                    e.Graphics.DrawLine(DepthBar.ProbePen, depthProbeVisualPos.X - _probeRadius, depthProbeVisualPos.Y - _probeRadius, depthProbeVisualPos.X + _probeRadius, depthProbeVisualPos.Y + _probeRadius);
-                    e.Graphics.DrawLine(DepthBar.ProbePen, depthProbeVisualPos.X - _probeRadius, depthProbeVisualPos.Y + _probeRadius, depthProbeVisualPos.X + _probeRadius, depthProbeVisualPos.Y - _probeRadius);
-                    e.Graphics.DrawString(i.ToString(), DepthBar.ProbeFont, DepthBar.ProbePen.Brush, new RectangleF(depthProbeVisualPos.X - 500.0f, depthProbeVisualPos.Y - 500.0f, 1000.0f, 500.0f), DepthBar.ProbeStringLayout);
+                    PointF depthProbeVisualPos = ToVisualCoord(_depthBar.DepthProbes[i].Position);
+
+                    RectangleF depthProbeRect = new RectangleF(depthProbeVisualPos.X - _probeRadius / 2, depthProbeVisualPos.Y - _probeRadius / 2, _probeRadius, _probeRadius);
+
+                    e.Graphics.SmoothingMode = SmoothingMode.AntiAlias;
+
+                    var probePen = new Pen(_depthBar.DepthProbes[i].Color, 2);
```
ActuallyaDeviloper: Drawing objects must be wrapped into a using block to make sure they are probably cleaned up.

 (btw can we use a few less empty lines here?)

TombEditor/Controls/DepthBar.cs
```diff
@@ -262,21 +314,61 @@ public void Draw(PaintEventArgs e, Size parentControlSize, Level level, Vector2
...
+                else if (distance > 0.0f)
+                {
+                    var alphaOutlinePen = (Pen)_outlinePen.Clone();
+                    alphaOutlinePen.Color = Color.FromArgb((int)(alphaOutlinePen.Color.A * (distance / _heightStringFadeDistance)), alphaOutlinePen.Color.R, alphaOutlinePen.Color.G, alphaOutlinePen.Color.B);
+
+                    DrawHeightString(e, barArea, alphaOutlinePen, (float)Math.Round(depth));
+
+                    if (i > 0 && i < _heightStringCount)
+                    {
+                        var alphaPen = (Pen)_heightLinesBigPen.Clone();
```
ActuallyaDeviloper: also here

TombEditor/Controls/DepthBar.cs
```diff
@@ -262,21 +314,61 @@ public void Draw(PaintEventArgs e, Size parentControlSize, Level level, Vector2
...
-                    DrawHeightString(e, barArea, _outlinePen, depth);
-                    e.Graphics.DrawLine(_heightLinesBigPen, barArea.Left, posY, barArea.Right, posY);
+                    DrawHeightString(e, barArea, _outlinePen, (float)Math.Round(depth));
+
+                    if(i > 0 && i < _heightStringCount)
+                        e.Graphics.DrawLine(_heightLinesBigPen, barArea.Left, posY, barArea.Right, posY);
+                }
+                else if (distance > 0.0f)
+                {
+                    var alphaOutlinePen = (Pen)_outlinePen.Clone();
```
ActuallyaDeviloper: We should wrap System Drawing objects in using blocks to make sure that they are cleaned up properly.

TombEditor/Controls/DepthBar.cs
```diff
@@ -38,31 +40,47 @@ public class DepthBar
...
 
         public static readonly StringFormat ProbeStringLayout = new StringFormat() { Alignment = StringAlignment.Center, LineAlignment = StringAlignment.Far };
-        public static readonly Font ProbeFont = new Font("Arial", 18.0f, FontStyle.Regular, GraphicsUnit.Pixel);
-        public static readonly Pen ProbePen = new Pen(Color.FromArgb(120, 0, 120), 2);
-        private static readonly Font _heightStringFont = new Font("Arial", 12.0f, FontStyle.Regular, GraphicsUnit.Pixel);
+        public static readonly Font ProbeFont = new Font("Segoe UI", 14.0f, FontStyle.Bold, GraphicsUnit.Pixel);
+        private static readonly Font _heightStringFont = new Font("Segoe UI", 11.0f, FontStyle.Bold, GraphicsUnit.Pixel);
         private static readonly StringFormat _heightStringLayout = new StringFormat() { Alignment = StringAlignment.Far, LineAlignment = StringAlignment.Center };
-        private static readonly Font _explainationStringFont = _heightStringFont;
+        private static readonly Font _explainationStringFont = new Font("Segoe UI", 12.0f, FontStyle.Regular, GraphicsUnit.Pixel);
```
ActuallyaDeviloper: not your fault, but can you please fix this embarrassing spelling of 'explanation'?

TombEditor/Controls/Panel2DMap.cs
Lwmte: Sorry, I usually prefer to divide code parts by their logic using empty lines =) I'll use less of them.

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -111,6 +111,23 @@ private void MoveToFixedPoint(PointF visualPoint, Vector2 worldPoint)
...
 
+        private int? FindClosestProbe(Vector2 clickPos)
+        {
+            if (_depthBar.DepthProbes.Count == 0)
+                return null;
+
+            int currentProbeIndex = 0;
+
+            for (int i = 0; i < _depthBar.DepthProbes.Count; ++i)
+                if ((clickPos - _depthBar.DepthProbes[i].Position).LengthSquared() < (clickPos - _depthBar.DepthProbes[currentProbeIndex].Position).LengthSquared())
```
ActuallyaDeviloper: I believe there is a bug here, I think it should be "Length" now instead of "LengthSquared".

PR #131, opened 2017-10-04
Implement selectable texturing

Lwmte:
Implements Photoshop-like (or Dxtre3d-like) selectable texturing. For this, in texturing mode, selection can be done with Ctrl+shift modifier. Note that selection doesn't display on textured tiles, most likely room shader must be modified for this.

Additionally, this PR implements disabling selection on "clicking away" on any empty space, and fixes some buttons in tool strip.


TombEditor/EditorActions.cs
```diff
@@ -922,42 +922,52 @@ public static void ApplyTextureAutomatically(Room room, DrawingPoint pos, BlockF
             _editor.RoomTextureChange(room);
         }
 
-        public static void TexturizeAllFloor(Room room, TextureArea texture)
+        public static void TexturizeAll(Room room, Rectangle area, TextureArea texture, BlockFaceType type)
```
ActuallyaDeviloper, line 930: I am not really sold on the introduction of this `BlockFaceType` enumeration, the code seems more intertwined and longer than just keeping them in there own function.

TombEditor/ToolWindows/MainView.cs
```diff
@@ -67,13 +67,16 @@ private void EditorEventRaised(IEditorEvent obj)
                 butCopy.Enabled = selectedObject is PositionBasedObjectInstance;
                 butStamp.Enabled = selectedObject is PositionBasedObjectInstance;
             }
+
             if (obj is Editor.SelectedSectorsChangedEvent)
```
ActuallyaDeviloper: Dead code now, please remove

TombEditor/ToolWindows/MainView.cs
```diff
@@ -91,6 +94,10 @@ private void EditorEventRaised(IEditorEvent obj)
 
                 panel2DMap.Visible = mode == EditorMode.Map2D;
                 panel3D.Visible = (mode == EditorMode.FaceEdit) || (mode == EditorMode.Geometry) || (mode == EditorMode.Lighting);
+
+                butTextureFloor.Enabled = mode == EditorMode.FaceEdit;
+                butTextureCeiling.Enabled = mode == EditorMode.FaceEdit;
+                butTextureWalls.Enabled = mode == EditorMode.FaceEdit;
```
ActuallyaDeviloper, line 100: Do we really want this? Just because you are not in texturing mode, the operation still makes perfectly sense to me.

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -787,7 +799,10 @@ protected override void OnMouseMove(MouseEventArgs e)
                         PickingResultBlock newPicking = DoPicking(GetRay(e.X, e.Y)) as PickingResultBlock;
 
                         if (newPicking != null)
-                            EditorActions.ApplyTextureAutomatically(_editor.SelectedRoom, newPicking.Pos, newPicking.Face, _editor.SelectedTexture);
+                        {
+                            if ((_editor.SelectedSectors.Valid && _editor.SelectedSectors.Area.Contains(newPicking.Pos) || _editor.SelectedSectors == SectorSelection.None))
```
ActuallyaDeviloper, line 794: Maybe I am overlooking something but this seems tautological to me to:
`!_editor.SelectedSectors.Valid || _editor.SelectedSectors.Area.Contains(newPicking.Pos)`

TombEditor/ToolWindows/MainView.cs
```diff
@@ -178,17 +181,17 @@ private void butOpacityTraversableFaces_Click(object sender, EventArgs e)
 
         private void butTextureFloor_Click(object sender, EventArgs e)
         {
-            EditorActions.TexturizeAllFloor(_editor.SelectedRoom, _editor.SelectedTexture);
+            EditorActions.TexturizeAllFloor(_editor.SelectedRoom, _editor.SelectedSectors.Area, _editor.SelectedTexture);
```
ActuallyaDeviloper, line 184: I think you should pass `_editor.SelectedSectors.Valid ? _editor.SelectedSectors.Area : _editor.SelectedRoom.LocalArea` because:
 - selection behaviour is abstracted outside the EditorAction function (If i pass an empty area to a function, I would find it surprising if it actually does everything)
 - `Area.Width == 0 && Area.Height == 0` also catches 1² sector selections (both coordinates are inclusive!), you should check `Valid` of the block selection instead.

TombEditor/ToolWindows/MainView.cs
Lwmte: It does not make sense, because user can't see result of the operation in current room.

TombEditor/EditorActions.cs
Lwmte: You mean what? I should use numerical constants instead of enum? Or I must bring back three Texturize* function, in which code was copy-pasted three times?

TombEditor/ToolWindows/MainView.cs
ActuallyaDeviloper: Maybe he does not need immediate feedback.
Maybe he does see it because he is eg in lighted mode.

Lwmte: Well, I always thought that according to WYSIWYG concept, user should be able to see changes that he made. Otherwise, from basic principle of human psychology, one may think that no visible action happened, and continue with other tasks.

Imagine this scenario: user edits level geometry and accidentally pushes "Texture all" button (or accidentally presses keycombo for that action). Since he's in geometry mode, no visible change in currently edited room has happened. User continues to edit level geometry, instinctively hitting Ctrl+S every 30 seconds, because he is a man frustrated by constant winroomedit crashes. Then, after some time, he enters face edit and finds out he accidentally textured all floor/ceiling/walls. Since no undo or backup system is implemented, he can't go back and revert his changes.

I agree on lighting mode, though. We can enable texturing commands in this mode as well, if you say it's needed.

PR #137, opened 2017-10-10
Easy item placement

Lwmte:
Allows dragging and dropping items from Object Preview right to 3D view.
Also allows to drag-n-drop projects, wads, texture sets and custom geometry right into editor window.

### DIFF
--- a/TombEditor/Controls/PanelRendering3D.cs
+++ b/TombEditor/Controls/PanelRendering3D.cs
@@ -96,15 +96,13 @@ namespace TombEditor.Controls
         {
             public DrawingPoint Pos { get; set; }
             public BlockFace Face { get; set; }
-            public bool IsFloorHorizontalPlane { get; private set; }
-            public bool BelongsToFloor { get; private set; }
+            public bool IsFloorHorizontalPlane => (Face == BlockFace.Floor || Face == BlockFace.FloorTriangle2);
+            public bool BelongsToFloor => (IsFloorHorizontalPlane || Face <= BlockFace.DiagonalMiddle);
             public PickingResultBlock(float distance, DrawingPoint pos, BlockFace face)
             {
                 Distance = distance;
                 Pos = pos;
                 Face = face;
-                IsFloorHorizontalPlane = (Face == BlockFace.Floor || Face == BlockFace.FloorTriangle2);
-                BelongsToFloor = (IsFloorHorizontalPlane || Face <= BlockFace.DiagonalMiddle);
             }
         }
 


TombEditor/Geometry/ImportedGeometry.cs
```diff
@@ -223,57 +224,95 @@ private Texture GetOrAddTexture(Dictionary<string, Texture> absolutePathTextureL
             return newTexture;
         }
 
-        public static string SupportedFormats
+        private struct FileFormat
         {
-            get
+            public string Description;
+            public string Extension;
```
ActuallyaDeviloper: I think we should make this an array, some file formats have a multitude of standard extensions.

TombEditor/Geometry/ImportedGeometry.cs
```diff
@@ -223,57 +224,95 @@ private Texture GetOrAddTexture(Dictionary<string, Texture> absolutePathTextureL
...
+                new FileFormat() { Description = "Neutral File Format (*.nff)", Extension = "nff" },
+                new FileFormat() { Description = "Object File Format (*.off)", Extension = "off" },
+                new FileFormat() { Description = "PovRAY Raw (*.raw)", Extension = "raw" },
+                new FileFormat() { Description = "Terragen Terrain (*.ter)", Extension = "ter" },
+                new FileFormat() { Description = "3D GameStudio(3DGS) (*.mdl)", Extension = "mdl" },
+                new FileFormat() { Description = "3D GameStudio(3DGS) Terrain (*.hmp)", Extension = "hmp" },
+                new FileFormat() { Description = "Izware Nendo (*.ndo)", Extension = "ndo" }
+            };
+
+            public static string GetFilter()
```
ActuallyaDeviloper: If we move it into TombLib, we could give it the `List<FileFormat>` as a parameter.

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -721,9 +723,21 @@ protected override void OnMouseDoubleClick(MouseEventArgs e)
...
+                    PickingResult newPicking = DoPicking(GetRay(e.X, e.Y));
+                    if (newPicking is PickingResultObject)
+                        EditorActions.EditObject(((PickingResultObject)newPicking).ObjectInstance, this.Parent);
+                    break;
+
+                case MouseButtons.Right:
+                    _editor.ResetCamera();
+                    break;
+
+                default:
```
ActuallyaDeviloper: This default is superfluous

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -805,6 +819,72 @@ protected override void OnMouseUp(MouseEventArgs e)
...
+                        info.Name = Path.GetFileNameWithoutExtension(files[i]);
+
+                        var instance = new ImportedGeometryInstance();
+                        var existingGeometry = _editor.Level.Settings.ImportedGeometries.Find(item => item.Info.Path == info.Path);
+
+                        if(existingGeometry == null)
+                        {
+                            existingGeometry = new ImportedGeometry();
+                            _editor.Level.Settings.ImportedGeometryUpdate(existingGeometry, info);
+                            _editor.Level.Settings.ImportedGeometries.Add(existingGeometry);
```
ActuallyaDeviloper: You should call `_editor.LoadedImportedGeometriesChange()` here

TombEditor/FormMain.cs
```diff
@@ -1031,5 +1031,16 @@ private void debugAction5ToolStripMenuItem_Click(object sender, EventArgs e)
         {
             NGTriggersDefinitions.LoadTriggers(File.OpenRead("NG\\NG_Constants.txt"));
         }
+
+        private void FormMain_DragDrop(object sender, DragEventArgs e)
```
ActuallyaDeviloper: Why not overriding here too?

TombEditor/EditorActions.cs
```diff
@@ -1846,6 +1846,82 @@ public static void Clone(IWin32Window owner)
...
+
+                        _editor.Level.Settings.WadFilePath = lowerFile;
+                        _editor.Level.ReloadWad();
+                        _editor.LoadedWadsChange(_editor.Level.Wad);
+                    }
+                    else if (lowerFile.EndsWith("png") || (lowerFile.EndsWith("tga")))
+                    {
+                        fileCount--;
+
+                        if (_editor.Level.Settings.TextureFilePath == lowerFile)
```
ActuallyaDeviloper: Currently you are relying on absolute paths everywhere. You should use `editor.Level.Settings.MakeRelative("your path", VariableType.LevelDirectory)` when assigning to the level paths, and `editor.Level.Settings.MakeAbsolute("your path")` when reading from them like in this comparison. Finally the comparison should be case insensitive because it's for windows.

This also applies to the wad and geometry import code.

TombEditor/EditorActions.cs
```diff
@@ -1846,6 +1846,82 @@ public static void Clone(IWin32Window owner)
...
+                            break;
+
+                        _editor.Level.Settings.TextureFilePath = lowerFile;
+                        _editor.LoadedTexturesChange();
+                    }
+                    else if (lowerFile.EndsWith("prj"))
+                    {
+                        OpenLevelPrj(lowerFile, owner);
```
ActuallyaDeviloper: I think it's dangerous to omit the "unsaved changes" dialog, don't you think so?

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -806,6 +817,75 @@ protected override void OnMouseUp(MouseEventArgs e)
...
+                    {
+                        if (!SupportedFormats.IsExtensionPresent(FileFormatType.Geometry, file))
+                            continue;
+
+                        var info = ImportedGeometryInfo.Default;
+                        info.Path = _editor.Level.Settings.MakeRelative(file, VariableType.LevelDirectory);
+                        info.Name = Path.GetFileNameWithoutExtension(file);
+
+                        var instance = new ImportedGeometryInstance();
+                        var geometryToDrop = _editor.Level.Settings.ImportedGeometries.Find(item => item.Info.Path == info.Path);
```
ActuallyaDeviloper: You are still comparing case sensitively. Also we really should compare the absolute paths instead of the relative path, because relative paths are more ambigous: There are many more ways to express a relative path to the same file.

TombLib/IO/FileFormat.cs
```diff
@@ -0,0 +1,199 @@
...
+
+    public static class SupportedFormats
+    {
+        private struct FileFormat
+        {
+            public string Description;
+            public List<string> Extensions;
+        }
+
+        private static List<FileFormat> _supportedGeometryFormats = new List<FileFormat>();
```
ActuallyaDeviloper, line 23: I think a more _functional style_ would be simpler and a _decentralized_ approach where TombLib does not dictate the file formats would easier to reuse and extend.

Maybe we should avoid unnecessary churn now though since now, it's already done in that way.

TombEditor/EditorActions.cs
```diff
@@ -1846,6 +1846,65 @@ public static void Clone(IWin32Window owner)
...
+                    {
+                        OpenLevelPrj(owner, file);
+                    }
+                    else if (file.EndsWith(".prj2", StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        OpenLevel(owner, file);
+                    }
+                }
+
+                return (fileCount == 0);
```
ActuallyaDeviloper: This will returns `false` for prj2 files. Is this indended? It looks wrong.

TombEditor/Controls/PanelRendering3D.cs
Lwmte: Probably I understood you wrong, but you said I must make all absolute paths relative?

TombLib/IO/FileFormat.cs
Lwmte: Well, generally I don't think that any other file type groups will be needed for editor? Except maybe for sound samples. I thought that we can make list of lists of file formats that would be referenced by string ID, but I decided not to implement this approach exactly because generally there are three file type groups anybody would ever use - wads, textures, custom models. Unless we're planning to turn Tomb Editor into full multimedia and 3D editor :D

TombEditor/EditorActions.cs
Lwmte: Oh, sorry, my bad. Actually this function could be made more simple just by inverting conditions.

TombEditor/Controls/PanelRendering3D.cs
ActuallyaDeviloper: I was referring to the `item.Info.Path == info.Path` comparsion. This comparsion should be done on absolute paths and case insensitively since they are not unique.

Example:
After I drop in the file I customize the path manually since it's actually, let's say, relative to the editor. `item.Info.Path` is now `$(EditorDirectory)/Something/TheFile.fbx`, but if I drop the file in again, `MakeRelative` will spit out eg `$(LevelDirectory)/../../../SomethingElse/Something/TheFile.fbx` again and the relative path comparison will result in `false` even though it's the same file.
I don't think this is the intended behaviour.

TombLib/IO/FileFormat.cs
ActuallyaDeviloper: Well the exact supported file types are a moving target depending on what versions of what libraries we are currently using. (eg Assimp is [adding new ones](https://github.com/assimp/assimp/blob/master/CHANGES) all the time)

Regarding full multimedia I actually thought about importing *.gif animations or small videos directly as animated textures. :P

Lwmte: "I actually thought about importing *.gif animations or small videos directly as animated textures." -- some day you will positively kill TRLE community with your ideas %) No, really, it's actually not that good idea too, cause gif files are 256-color only, which immediately will be seen in game, and videos are generally compressed too (except maybe lossless codecs, but supporting those will be a burden anyway).

Ok, if we come to point when file format groups are in need to be increased, I will redo this code with more flexible approach :)

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -94,13 +96,15 @@ private class PickingResultBlock : PickingResult
         {
             public DrawingPoint Pos { get; set; }
             public BlockFace Face { get; set; }
-            public bool IsFloor { get; private set; }
+            public bool IsFloorHorizontalPlane { get; private set; }
```
ActuallyaDeviloper:
```
public bool IsFloorHorizontalPlane => (Face == BlockFace.Floor || Face == BlockFace.FloorTriangle2);
public bool BelongsToFloor => (IsFloorHorizontalPlane || Face <= BlockFace.DiagonalMiddle);
```
would be nice to avoid state.

PR #146, opened 2017-10-12
Minor cleanups and GUI tweaks

Banderi:
Changelog:
- made cardinal points text more centered
- made camera pan and zoom speed reflect camera zoom (distance)
- made camera follow current room on room displacement
- *SLIGHTLY* tweaked default navigation speeds
- inverted click-and-drag zooming direction by default to reflect standard behavior (http://www.tombraiderforums.com/showthread.php?p=7787278&posted=1#post7787278)
- fixed Bounding Box-only ray-picking for imported geometry
- made so void left clicking (clicking outside room) deselects any tiles and objects, and selecting tiles or objects deselects any of the other (behavior can be overridden with CTRL)
- tweaked the 2D View keymappings
- changed the default **YesNo** messagebox for file closing to more common **YesNoCancel** box that asks if you want to save unsaved changes, or not (also made into global method)
- added basic flag for unsaved changes in the editor, which updates the title bar to reflect current status (e.g. _"test"_ is saved, _"test*"_ has unsaved changes); the basic flag could be swapped later on for undo/redo population
- made the primary save button (CTRL+S) not do anything if there are no unsaved changes (temporarily tweaked the WS tile editing keys as well to reflect this)
- switched `FolderBrowserDialog()` method for `CommonOpenFileDialog()` using the WindowsAPICodePack package (now added to NuGet)


TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -557,6 +568,10 @@ protected override void OnMouseDown(MouseEventArgs e)
                 {
                     _editor.SelectedObject = ((PickingResultObject)newPicking).ObjectInstance;
                 }
+                else if (!(newPicking is PickingResultGizmo))
+                {
+                    _editor.SelectedObject = null;
+                }
```
Lwmte: I did it in a bit another way, check it out here: https://github.com/MontyTRC89/Tomb-Editor/blob/8855302581f83288c90c0f9c50c739b582d877f2/TombEditor/Controls/PanelRendering3D.cs#L564

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -693,6 +708,10 @@ protected override void OnMouseDown(MouseEventArgs e)
                                         _doSectorSelection = true;
                                     }
                                 }
+                                else
+                                {
+                                    _editor.SelectedSectors = SectorSelection.None;
```
Lwmte: Is this needed at all? In my case it worked without this line.

TombEditor/FormMain.cs
```diff
@@ -182,12 +183,21 @@ private void EditorEventRaised(IEditorEvent obj)
...
+                if (_editor.UnsavedChanges)
+                {
+                    Text = "Tomb Editor " + Application.ProductVersion.ToString() + " - " + LevelName + "*";
+                    saveLevelToolStripMenuItem.Enabled = true;
+                }
+                else
+                {
+                    Text = "Tomb Editor " + Application.ProductVersion.ToString() + " - " + LevelName;
+                    saveLevelToolStripMenuItem.Enabled = false;
+                }
```
Lwmte: I think you can simplify code like this:

                Text = "Tomb Editor " + Application.ProductVersion.ToString() + " - " + LevelName;
                saveLevelToolStripMenuItem.Enabled = false;
                
                if (_editor.UnsavedChanges)
                {
                    Text = Text + "*";
                    saveLevelToolStripMenuItem.Enabled = true;
                }

Banderi: Very true, I'll fix! (although that double flag set makes me die a bit inside...)

TombEditor/Controls/PanelRendering3D.cs
Banderi: If I remove that line tiles don't get deselected by clicking outside room - maybe you have redundancy from your own code?

TombEditor/Controls/DepthBar.cs
```diff
@@ -84,7 +84,7 @@ public DepthProbe(DepthBar parent)
                 Color = parent.getProbeColor();
             }
         }
-        
+
```
ActuallyaDeviloper, line 87: Thanks for cleaning up the trailing white spaces btw

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -166,46 +189,54 @@ protected override void OnMouseDown(MouseEventArgs e)
...
+                            _currentlyEditedDepthProbeIndex = _depthBar.DepthProbes.Count;
+                            _depthBar.DepthProbes.Add(new DepthBar.DepthProbe(_depthBar));
+                            _depthBar.DepthProbes[_currentlyEditedDepthProbeIndex.Value].Position = clickPos;
+                        }
                         Invalidate();
                     }
                     break;
+                case MouseButtons.Right:
+                case MouseButtons.XButton2:
+                    break;
```
ActuallyaDeviloper: Those cases are pointless

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -166,46 +189,54 @@ protected override void OnMouseDown(MouseEventArgs e)
...
                     if (_currentlyEditedDepthProbeIndex.HasValue)
                     {
-                        // Remove depth probe closest to mouse pointer
                         _depthBar.DepthProbes.RemoveAt(_currentlyEditedDepthProbeIndex.Value);
                         Invalidate();
                     }
                     break;
+
+                case MouseButtons.XButton1:
+                    break;
```
ActuallyaDeviloper: This case is pointless

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -224,11 +227,18 @@ private void EditorEventRaised(IEditorEvent obj)
                 (obj is Editor.LoadedTexturesChangedEvent) ||
                 (obj is Editor.LoadedImportedGeometriesChangedEvent))
             {
+                if (Camera != null && (obj is Editor.SelectedRoomChangedEvent || obj is Editor.ModeChangedEvent))
```
ActuallyaDeviloper: I am not quite getting the purpose of this code. Can you explain?

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -547,22 +558,32 @@ protected override void OnMouseDown(MouseEventArgs e)
...
+
+                if (!(newPicking is PickingResultBlock))
+                {
+                    // No tile selected
+                    _editor.SelectedSectors = SectorSelection.None;
+                }
+                if (!(newPicking is PickingResultGizmo) && !(newPicking is PickingResultObject))
+                {
+                    // No object or object gizmo selected
+                    _editor.SelectedObject = null;
```
ActuallyaDeviloper: I am quite honest, I don't like this very strict version of unselect on clicking outside. I think this is both inconvinient and inconsistent. All other ways to select objects and sectors (2D grid, trigger list, many tools) do not behave this way.

Now I can now longer work on a light and sectors at the same time except if I am careful enough to select through the 2D grid, but this is inconvinient.

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -1531,7 +1555,7 @@ private void DrawMoveables(Matrix viewProjection)
 
                 if (_editor.SelectedObject == instance)
                 {
-                    Vector3 screenPos = Vector3.Project(512.0f * Vector3.UnitY, 0, 0, Width,
+                    Vector3 screenPos = Vector3.Project(512.0f * Vector3.UnitY, -10, 40, Width,
```
ActuallyaDeviloper: This change seems very weird to me?
What are you trying to achieve? An accident?

TombEditor/EditorActions.cs
```diff
@@ -1888,13 +1906,16 @@ public static void SaveLevel(IWin32Window owner, bool askForPath)
...
+                    _editor.UnsavedChanges = false;
                     _editor.LevelFileNameChange();
                 }
 
             // Save level
             try
             {
                 Prj2Writer.SaveToPrj2(_editor.Level.Settings.LevelFilePath, _editor.Level);
+                _editor.UnsavedChanges = false;
+                _editor.LevelFileNameChange();
```
ActuallyaDeviloper: `LevelFileNameChange` should not be called here, if name actually changed it was already called above.

TombEditor/EditorActions.cs
```diff
@@ -1888,13 +1906,16 @@ public static void SaveLevel(IWin32Window owner, bool askForPath)
                         return;
 
                     _editor.Level.Settings.LevelFilePath = saveFileDialog.FileName;
+                    _editor.UnsavedChanges = false;
```
ActuallyaDeviloper: This should not be set here, it is already set below unconditionally

TombEditor/FormLevelSettings.cs
```diff
@@ -267,16 +268,27 @@ private string BrowseFile(string currentPath, string title, string filter, Varia
...
-            {
-                dialog.Description = description;
-                dialog.SelectedPath = _levelSettings.MakeAbsolute(currentPath);
-                dialog.RootFolder = Environment.SpecialFolder.MyComputer;
-                SendKeys.Send("{TAB}{TAB}{RIGHT}");  // Workaround https://stackoverflow.com/questions/6942150/why-folderbrowserdialog-dialog-does-not-scroll-to-selected-folder
-                if (dialog.ShowDialog(this) != DialogResult.OK)
-                    return null;
-                return _levelSettings.MakeRelative(dialog.SelectedPath, baseDirType);
-            }
+            if (true)
```
ActuallyaDeviloper: Can we remove the obsolete dialog code?

TombEditor/FormMain.cs
```diff
@@ -182,12 +183,20 @@ private void EditorEventRaised(IEditorEvent obj)
...
 
-            // Update application title
+            // Update application title and save button
             if (obj is Editor.LevelFileNameChanged)
             {
                 string LevelName = string.IsNullOrEmpty(_editor.Level.Settings.LevelFilePath) ? "Untitled" :
                     Path.GetFileNameWithoutExtension(_editor.Level.Settings.LevelFilePath);
+
                 Text = "Tomb Editor " + Application.ProductVersion.ToString() + " - " + LevelName;
+                saveLevelToolStripMenuItem.Enabled = false;
```
ActuallyaDeviloper: I think a different event should be introduced for unsaved changes

TombEditor/FormMain.cs
```diff
@@ -200,6 +209,20 @@ private void EditorEventRaised(IEditorEvent obj)
                     (@event.Current.Window_Layout != @event.Previous.Window_Layout))
                     LoadWindowLayout(_editor.Configuration);
             }
+
+            // Update on changes to the project
+            if (((obj is IEditorObjectChangedEvent) && !(obj is Editor.SelectedObjectChangedEvent)) ||
```
ActuallyaDeviloper: And this event code should be moved to the editor since it actually does not update the main form but "UnsavedChanges" of the editor

TombEditor/TombEditor.csproj
```diff
@@ -65,6 +65,26 @@
...
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Microsoft.WindowsAPICodePack.Shell, Version=1.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\packages\WindowsAPICodePack.1.1.0\lib\Microsoft.WindowsAPICodePack.Shell.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Microsoft.WindowsAPICodePack.ShellExtensions, Version=1.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\packages\WindowsAPICodePack.1.1.0\lib\Microsoft.WindowsAPICodePack.ShellExtensions.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
```
ActuallyaDeviloper: What are those new libraries for? Do we really need them?

TombEditor/ToolWindows/Lighting.cs
```diff
@@ -39,7 +39,7 @@ private void EditorEventRaised(IEditorEvent obj)
         {
             // Update light UI
             if ((obj is Editor.ObjectChangedEvent) ||
-               (obj is Editor.SelectedObjectChangedEvent))
+               (obj is Editor.ObjectChangedEvent))
```
ActuallyaDeviloper: Accidental change? Light panel update is now broken.

TombLib/Graphics/BaseGizmo.cs
```diff
@@ -55,13 +55,15 @@ public abstract class BaseGizmo : IDisposable
         private readonly GeometricPrimitive _cylinder;
         private readonly GeometricPrimitive _sphere;
```
ActuallyaDeviloper: Please remove `_sphere`, we no longer use it

TombLib/Graphics/BaseGizmo.cs
```diff
@@ -401,27 +405,27 @@ public void Draw(Matrix viewProjection)
                 }
                 _torus.SetupForRendering(_device);
 
-                // Rotation Y
-                if (SupportRotationY)
+                // Rotation X
+                if (SupportRotationX)
```
ActuallyaDeviloper: Y was always put first because the rotations are applied in the Y, X, Z order

TombLib/Graphics/BaseGizmo.cs
```diff
@@ -441,10 +445,51 @@ public void Draw(Matrix viewProjection)
                 }
             }
 
+            _device.Clear(ClearOptions.DepthBuffer, Color4.Black, 1.0f, 0);
```
ActuallyaDeviloper: What is the goal with this? Now the geometry is no longer ordered and using the gizmo makes me more nauseous again because things that are in the same space are not intersect like they would in a real world.

TombLib/Graphics/BaseGizmo.cs
```diff
@@ -441,10 +445,51 @@ public void Draw(Matrix viewProjection)
...
+
             // Scale
             if (SupportScale)
             {
+                _cylinder.SetupForRendering(_device);
+                _device.SetRasterizerState(_device.RasterizerStates.CullFront);
+
```
ActuallyaDeviloper: CullFront is already set, this is unnecessary

TombLib/Graphics/BaseGizmo.cs
```diff
@@ -441,10 +445,51 @@ public void Draw(Matrix viewProjection)
...
+                        Matrix.RotationX(-(float)Math.PI / 2.0f) *
+                        Matrix.Translation(Position);
+                    solidEffect.Parameters["ModelViewProjection"].SetValue(model * viewProjection);
+                    solidEffect.Parameters["Color"].SetValue(_zAxisColor + (highlight == GizmoMode.ScaleZ ? _hoveredAddition : new Color4()));
+                    solidEffect.CurrentTechnique.Passes[0].Apply();
+                    _device.DrawIndexed(PrimitiveType.TriangleList, _cylinder.IndexBuffer.ElementCount);
+                }
+
                 _cube.SetupForRendering(_device);
+                _device.SetRasterizerState(_device.RasterizerStates.CullBack);
```
ActuallyaDeviloper: unnecessary

TombLib/Graphics/BaseGizmo.cs
```diff
@@ -481,6 +526,7 @@ public void Draw(Matrix viewProjection)
             if (SupportTranslate)
             {
                 _cylinder.SetupForRendering(_device);
+                _device.SetRasterizerState(_device.RasterizerStates.CullFront);
```
ActuallyaDeviloper: unnecessary

TombLib/Graphics/BaseGizmo.cs
```diff
@@ -512,41 +558,44 @@ public void Draw(Matrix viewProjection)
...
-                    solidEffect.Parameters["Color"].SetValue(_yAxisColor + (highlight == GizmoMode.TranslateY ? _hoveredAddition : new Color4()));
+                    solidEffect.Parameters["Color"].SetValue(_yAxisColor + (highlight == GizmoMode.TranslateY ? _hoveredAddition * _hoveredAdditionGreensCorrection : new Color4()));
                     solidEffect.CurrentTechnique.Passes[0].Apply();
-                    _device.DrawIndexed(PrimitiveType.TriangleList, _sphere.IndexBuffer.ElementCount);
+                    _device.DrawIndexed(PrimitiveType.TriangleList, _cone.IndexBuffer.ElementCount);
                 }
 
                 // Z axis translation
                 {
                     var model = Matrix.Scaling(TranslationSphereSize) *
```
ActuallyaDeviloper: I think we should rename `TranslationSphereSize` to `TranslationConeSize` everywhere

TombLib/Graphics/BaseGizmo.cs
```diff
@@ -246,15 +250,15 @@ public PickingResultGizmo DoPicking(Ray ray)
             if (SupportTranslate)
             {
                 float unused;
-                BoundingSphere sphereX = new BoundingSphere(Position + Vector3.UnitX * Size, TranslationSphereSize / 2.0f);
```
ActuallyaDeviloper, line 249: Can we give this proper collision checks for cones?
http://www.realtimerendering.com/intersections.html

TombEditor/EditorActions.cs
```diff
@@ -1955,6 +1973,7 @@ public static void OpenLevelPrj(IWin32Window owner)
                     {
                         if (form.ShowDialog(owner) != DialogResult.OK || newLevel == null)
                             return;
+                        _editor.UnsavedChanges = false;
```
ActuallyaDeviloper: It is already updated directly inside the Editor.Level setter, we don't need this here.

TombEditor/EditorActions.cs
```diff
@@ -22,6 +22,24 @@ public static class EditorActions
 
         private static Editor _editor = Editor.Instance;
 
+        public static bool ContinueOnFileDrop(IWin32Window owner, string description)
+        {
+            if (!_editor.UnsavedChanges)
+                return true;
+
+            DialogResult saveChanges = DarkMessageBox.Show(owner,
```
ActuallyaDeviloper: switch case maybe?

TombEditor/EditorActions.cs
```diff
@@ -1928,15 +1947,14 @@ public static void OpenLevel(IWin32Window owner)
                     logger.Error(exc, "Unable to open \"" + openFileDialog.FileName + "\"");
                     DarkMessageBox.Show(owner, "There was an error while opening project file. File may be in use or may be corrupted. Exception: " + exc.Message, "Error", MessageBoxIcon.Error);
                 }
+                _editor.UnsavedChanges = false;
```
ActuallyaDeviloper: It is already updated directly inside the Editor.Level setter, we don't need this here.

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -148,14 +148,37 @@ protected override void OnMouseDown(MouseEventArgs e)
...
+                        else
+                        {
+                            // Add depth probe under mouse pointer
+                            _currentlyEditedDepthProbeIndex = _depthBar.DepthProbes.Count;
+                            _depthBar.DepthProbes.Add(new DepthBar.DepthProbe(_depthBar));
+                            _depthBar.DepthProbes[_currentlyEditedDepthProbeIndex.Value].Position = clickPos;
+                        }
+                        Invalidate();
+                    }
+                    else if ((ModifierKeys & Keys.Alt) == Keys.Alt && _currentlyEditedDepthProbeIndex.HasValue)
```
ActuallyaDeviloper: ModifierKeys.HasFlag(Keys.Alt)
but what is the purpose of this anyway?

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -148,14 +148,37 @@ protected override void OnMouseDown(MouseEventArgs e)
...
-                    _roomMouseClicked = DoPicking(clickPos);
-                    if (_roomMouseClicked == null)
-                        return;
-
-                    _editor.SelectRoomAndResetCamera(_roomMouseClicked);
-                    _roomsToMove = _editor.Level.GetConnectedRooms(_editor.SelectedRoom);
-                    _roomMouseOffset = clickPos - _roomMouseClicked.SectorPos.ToVec2();
+                    // Try selecting or moving a room or depth probe
+                    _currentlyEditedDepthProbeIndex = FindClosestProbe(clickPos);
+                    if ((ModifierKeys & Keys.Control) == Keys.Control)
```
ActuallyaDeviloper: ModifierKeys.HasFlag(Keys.Control)

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -222,17 +253,23 @@ protected override void OnMouseMove(MouseEventArgs e)
...
+                            Invalidate();
+                        }
+                        else if (_roomsToMove != null)
                             UpdateRoomPosition(FromVisualCoord(e.Location) - _roomMouseOffset, _roomMouseClicked, _roomsToMove);
                         break;
 
+                    case MouseButtons.Middle:
                     case MouseButtons.Right:
                         if (_viewMoveMouseWorldCoord != null)
                             if (ModifierKeys.HasFlag(Keys.Control))
```
ActuallyaDeviloper: I think we shouldn't require Ctrl if the middle mouse button is used

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -166,46 +189,54 @@ protected override void OnMouseDown(MouseEventArgs e)
 
                 case MouseButtons.Middle:
                 case MouseButtons.XButton2:
```
ActuallyaDeviloper: I think the back button (XButton1) makes more sense to remove depth probes than the forward button (XButton2).
I also think that it would make a lot of sense to keep the forward button (XButton2) for adding new depth probes.

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -166,46 +189,54 @@ protected override void OnMouseDown(MouseEventArgs e)
...
             switch (e.Button)
             {
-                case MouseButtons.Middle:
-                case MouseButtons.XButton2:
-                    Vector2 clickPos = FromVisualCoord(e.Location);
-                    int? currentProbeIndex = FindClosestProbe(clickPos);
-                    if (currentProbeIndex.HasValue)
+                case MouseButtons.Left:
+                    // Remove depth probe closest to mouse pointer
+                    if ((ModifierKeys & Keys.Control) != Keys.Control)
```
ActuallyaDeviloper: !ModifierKeys.HasFlag(Keys.Control)

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -148,14 +148,37 @@ protected override void OnMouseDown(MouseEventArgs e)
...
-                        return;
-
-                    _editor.SelectRoomAndResetCamera(_roomMouseClicked);
-                    _roomsToMove = _editor.Level.GetConnectedRooms(_editor.SelectedRoom);
-                    _roomMouseOffset = clickPos - _roomMouseClicked.SectorPos.ToVec2();
+                    // Try selecting or moving a room or depth probe
+                    _currentlyEditedDepthProbeIndex = FindClosestProbe(clickPos);
+                    if ((ModifierKeys & Keys.Control) == Keys.Control)
+                    {
+                        // Remove depth probe closest to mouse pointer
```
ActuallyaDeviloper: Please move this comment into the if, since it's wrong otherwise

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -148,14 +148,37 @@ protected override void OnMouseDown(MouseEventArgs e)
...
-                    // Try selecting or moving a room
-                    _roomMouseClicked = DoPicking(clickPos);
-                    if (_roomMouseClicked == null)
-                        return;
-
-                    _editor.SelectRoomAndResetCamera(_roomMouseClicked);
-                    _roomsToMove = _editor.Level.GetConnectedRooms(_editor.SelectedRoom);
-                    _roomMouseOffset = clickPos - _roomMouseClicked.SectorPos.ToVec2();
+                    // Try selecting or moving a room or depth probe
+                    _currentlyEditedDepthProbeIndex = FindClosestProbe(clickPos);
```
ActuallyaDeviloper: Settings this upfront causes an exception later if the last depth probe was removed.
We should find the depth probe in local varaible first and only set "+                    _currentlyEditedDepthProbeIndex" if it is going to be moved around.

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -166,46 +189,54 @@ protected override void OnMouseDown(MouseEventArgs e)
...
-                    _depthBar.DepthProbes[_currentlyEditedDepthProbeIndex.Value].Position = clickPos;
-                    Invalidate();
-                    break;
+                    // Move view with mouse curser
+                    // Mouse curser is a fixed point
+                    _viewMoveMouseWorldCoord = clickPos;
 
-                case MouseButtons.XButton1:
+                    // Remove depth probe closest to mouse pointer
                     _currentlyEditedDepthProbeIndex = FindClosestProbe(clickPos);
```
ActuallyaDeviloper: Same as above, this causes an exception in some cases. `_currentlyEditedDepthProbeIndex` is used for moving depth probes around.

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -166,46 +189,54 @@ protected override void OnMouseDown(MouseEventArgs e)
...
-                        _depthBar.DepthProbes.RemoveAt(currentProbeIndex.Value);
+                        int? currentProbeIndex = FindClosestProbe(clickPos);
+                        if (currentProbeIndex.HasValue)
+                        {
+                            _depthBar.DepthProbes.RemoveAt(currentProbeIndex.Value);
+                        }
+                        else
+                        {
+                            // Add depth probe under mouse pointer
+                            _currentlyEditedDepthProbeIndex = _depthBar.DepthProbes.Count;
```
ActuallyaDeviloper: Don't use _currentlyEditedDepthProbeIndex here

TombEditor/Controls/Panel2DMap.cs
```diff
@@ -247,7 +284,6 @@ protected override void OnMouseMove(MouseEventArgs e)
                             }
                         break;
 
-                    case MouseButtons.Middle:
                     case MouseButtons.XButton2:
```
ActuallyaDeviloper: I don't think it makes sense to move depth probes around by pressing the back button (XButton2).

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -527,8 +537,9 @@ protected override void OnPreviewKeyDown(PreviewKeyDownEventArgs e)
         protected override void OnMouseWheel(MouseEventArgs e)
         {
             base.OnMouseWheel(e);
+            float distanceMultiplier = Camera.Distance / DefaultCameraDistance;
 
-            Camera.Zoom(-e.Delta * _editor.Configuration.Rendering3D_NavigationSpeedMouseWheelZoom);
+            Camera.Zoom(-e.Delta * _editor.Configuration.Rendering3D_NavigationSpeedMouseWheelZoom * distanceMultiplier);
```
ActuallyaDeviloper: I think we should move this logic into the "Zoom" function itself. This way it also applies to all places automatically including the item view.

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -923,7 +944,10 @@ private PickingResult DoPicking(Ray ray)
                     BoundingBox box = geometry.Model?.DirectXModel?.BoundingBox ?? new BoundingBox(new Vector3(-128), new Vector3(128));
```
ActuallyaDeviloper: not your fault, but we should use the `_littleCubeRadius` constant here.

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -923,7 +944,10 @@ private PickingResult DoPicking(Ray ray)
...
                     box.Minimum += room.WorldPos + instance.Position;
                     box.Maximum += room.WorldPos + instance.Position;
-                    if (ray.Intersects(ref box, out distance) && ((result == null) || (distance < result.Distance)))
+
+                    if (geometry?.Model?.DirectXModel?.Meshes?.ElementAt(0) != null)
```
ActuallyaDeviloper: We should use a foreach over all the meshes here like for the movable, otherwise:
 - imported geometry with more than 1 mesh is only partially selectable
 - imported geometry with 0 meshes causes a crash

Furthermore the bounding box stuff you are doing here is really weird. You called it `transformBB` but actually what it does is just disabling the bounding box check. This will affect the performance negatively for complex cases because the triangle by triangle case is always performed.

I think instead of this weird workaround with passing in `geometry.Model?.DirectXModel?.BoundingBox` and then ignoring it afterall, we should keep using the bounding box of the mesh as before. It won't work because `BoundingBox` is uninitialized for imported geometry just now, but we can just fix that issue instead. We just have to calculate the bounding box in `ImportedGeometryMesh.BuildBuffers`.

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -2224,7 +2248,7 @@ private void AddDirectionsToDebug(Matrix viewProjection)
                 Vector3 screenPos = Vector3.Project(positions[i], 0, 0, Width, Height,
                     _device.Viewport.MinDepth,
                     _device.Viewport.MaxDepth, wvp);
-                _debug.AddString(messages[i], screenPos);
+                _debug.AddString(messages[i], screenPos - new Vector3(45, 0, 0));       // it'd be better if it was a dynamic string length enumeration
```
ActuallyaDeviloper: What do you mean with a "dynamic string length enumeration"?

TombEditor/FormMain.cs
```diff
@@ -42,6 +42,7 @@ public FormMain(Editor editor)
             _editor.EditorEventRaised += EditorEventRaised;
 
             Text = "Tomb Editor " + Application.ProductVersion + " - Untitled";
+            //_editor.UnsavedChanges = true;
```
ActuallyaDeviloper: This comment appears to be not useful

TombLib/Graphics/BaseGizmo.cs
```diff
@@ -512,41 +558,44 @@ public void Draw(Matrix viewProjection)
...
+                _cone.SetupForRendering(_device);
+                _device.SetRasterizerState(_device.RasterizerStates.CullNone);
 
                 // X axis translation
                 {
-                    var model = Matrix.Scaling(TranslationSphereSize) *
-                        Matrix.Translation(Position + Vector3.UnitX * Size);
+                    var model = Matrix.RotationY((float)-Math.PI * 0.5f) *
+                        Matrix.Scaling(TranslationSphereSize) *
+                        Matrix.Translation(Position + (Vector3.UnitX + new Vector3(0.1f, 0, 0)) * Size);
```
ActuallyaDeviloper: new Vector3(1.1f, 0, 0) seems more on point to me. I would also suggest making 1.1 a private constant since its used in 3 places.

TombEditor/Controls/PanelRendering3D.cs
Banderi: This is so that the camera follows the active room when the room's position is changed in 2D View

TombEditor/Controls/PanelRendering3D.cs
Banderi: Could you elaborate more on the various use cases in which this would be inconvenient?
I've tweaked it a bit so that:
- you don't accidentally unselect stuff by simply clicking outside
- you can maintain object selection by using CTRL (wanted to add multiple tiles and object selection, but no way to do so for the time being... might look into that!)

TombEditor/Controls/PanelRendering3D.cs
Banderi: I've set the default bool parameter to `false` so unless it's specifically used the original BB will be used. I've made it like that so that imported geometry and other complex models can be selected precisely without using a simple averaged BB to approximate - and the original BB is used as a first-step check so you won't get degraded performance all the time; if I've misunderstood and there's a way to construct a collision object that still needs to be implemented, go for it!

I have also added a basic loop to test each mesh in the object as well if that was the first issue. I also assumed that the `?.` member selection "operator" was gonna safely return a `false` and therefore skip the offending loop; if that was wrong, my bad!

TombEditor/Controls/PanelRendering3D.cs
Banderi: I moved the debug info so it's slightly less in the way of the gizmo; not a perfect solution, but is a bit better than before.

TombEditor/Controls/PanelRendering3D.cs
Banderi: Actually in retrospect it's better to just have it moved by a fixed amount (no need to waste CPU cycles on such a nitpick) but what I meant with that was I couldn't find a way to enumerate string length and center the text on screen based on that. I'll remove that comment now!

TombEditor/EditorActions.cs
Banderi: What do you mean?

TombEditor/FormMain.cs
Banderi: I agree, but there's no event catch in Editor.cs or EditorActions.cs; as I said to Lwmte previously, there wasn't really a better module to stick this into that I could find.

TombEditor/TombEditor.csproj
Banderi: Yes, they are needed for the other folder dialog selection (as stated in the changelog)

TombEditor/FormMain.cs
Banderi: Yeah, that would be perfect to implement.

TombLib/Graphics/BaseGizmo.cs
Banderi: SharpDX does not seem to possess any easy ray-cone collision detection; the closest Bounding object is frustum, but I felt like it'd be better to keep the sphere until we can implement a proper cone collision.

TombLib/Graphics/BaseGizmo.cs
Banderi: Could you elaborate on the cases in which it happens? This is to make the rotational gizmo not overlap graphically with the translation gizmo, which is really ugly:

TombLib/Graphics/BaseGizmo.cs
Banderi: No, the initial culling state is `CullBack`. These subsequent alternating state changes are to make the gizmo graphics work correctly, so they're all needed (sans reworking internally how the faces are rendered with primitives).

TombLib/Graphics/BaseGizmo.cs
Banderi: May I ask why `Vector3.UnitX` (and equivalents) were used?

TombLib/Graphics/BaseGizmo.cs
Lwmte: Sphere actually is better IMO, because picking arrow on the tip will be too bothersome

Banderi: Well the cone has a pretty chunky shape, I kept it fairly large - cone collision would be as big as the sphere, with almost no noticeable difference - but if we did cone shape collision we'd definitely test to see which one seems better.

Lwmte: The trend is, "hard-to-pick" controls are currently created with a bit inflated click bounds, so if user misses a pixel of arrow tip, he will still be able to drag it. On the other hand, in case of arrows it may be better to use just bounding box (because arrows are long objects, if you enlarge sphere to fit it all, it may go beyond comfortable bounds above and below arrow). But not specifically cone shape.

TombLib/Graphics/BaseGizmo.cs
ActuallyaDeviloper: Well maybe it's ugly but it also makes sense. The none overlapping case makes my head hurt because it makes no sense geometrically.

TombLib/Graphics/BaseGizmo.cs
ActuallyaDeviloper: Ah I see now, sorry I did not catch that.

However instead of introducing competing standards we should turn the winding order of the cone shape itself which is almost certainly just an oversight because it was not used it. You can just swap the indices in `GeometricPrimitive.Cone.cs`.

TombEditor/FormMain.cs
ActuallyaDeviloper: We can just introduce them inside Editor.cs.

Just do `EditorEventRaised += Editor_EditorEventRaised` in the constructor and add:
```
private void Editor_EditorEventRaised(IEditorEvent obj)
{
    // The event code here
}
```

TombEditor/FormMain.cs
ActuallyaDeviloper: Let's do it then :)

TombEditor/EditorActions.cs
ActuallyaDeviloper:
```
switch (DarkMessageBox.Show(owner, "Your unsaved changes will be lost. Do you want to save?",
      description, MessageBoxButtons.YesNoCancel, MessageBoxIcon.Question))
{
    case DialogResult.No:
        return true;
    case DialogResult.Yes:
        return SaveLevel(owner, false); // By the way I just notice that this is dangerous, saving could fail, we should only continue if we haven't failed
    default:
        return false;
}
```

TombEditor/Controls/PanelRendering3D.cs
ActuallyaDeviloper: Ah I understand, this however is then inconsistent with many other debug strings.
If you search for `Vector3.Project` you will find a lot of other cases where this happens.

I am also not sure if this is a good way the move the text, but either way we should probably put inside a function and used everywhere, then it's easy to adjust too.

TombLib/Graphics/BaseGizmo.cs
Banderi: Actually that was my bad there, no need to set it back and forth. Cleaned!

TombEditor/FormMain.cs
Banderi: This should be done in a separate feature branch along with other editor events and undo/redo population, plus it's beyond my current capabilities

TombEditor/Controls/PanelRendering3D.cs
ActuallyaDeviloper: Maybe you intend to to do a bounding box check but actually you don't.
Currently it is implemented in such a way that no bounding box check is performed for imported geometry whatsoever.

TombEditor/EditorActions.cs
Banderi: SaveLevel() is a procedure, it should internally handling exceptions; I am not capable of catching failures outside of that function

TombEditor/FormMain.cs
ActuallyaDeviloper: I agree that in that in the long term this should be redone using redo/undo infrastructure.
Anyhow I think that we should convolute the "NeedsToBeSaved" state with "LevelFileNameChanged". Should be a simple change. If you want I can do it too.

TombEditor/Controls/PanelRendering3D.cs
ActuallyaDeviloper: I generally can't conviniently move an objects very quickly without losing by sector selection, I can't easily tweak sectors without losing object selection.

What do you think of Lwmte's suggestion: https://github.com/MontyTRC89/Tomb-Editor/pull/146/files/af023c5ba2d82634742a563efd9303dc38621692#diff-08f2f745e28a333b7eaa6b83b8294780R576
It only unselects stuff if you *really* click into the void, I think that's less intrusive.

TombEditor/Controls/PanelRendering3D.cs
ActuallyaDeviloper: Ah I see. I think you should move this outside this event block which is for invalidation and listen to `RoomGeometryChangedEvent` which will fire specifically at the time a room is moved.

Also `deltaRoomPos * 1024` is incorrect for the y-movment because y is in steps of `256`. Luckily you can just use `editor.SelectedRoom.WorldPos` which will handle all this for you.

TombEditor/Controls/Panel2DMap.cs
Lwmte: Can you explain why? Being able to move just created probe to correct its position is very handy.

TombEditor/Controls/Panel2DMap.cs
Lwmte: I can't find the code in question, is it still actual for current revision?

TombEditor/Controls/Panel2DMap.cs
Lwmte: Fixed as well, can't find in current revision

TombEditor/FormMain.cs
Lwmte: Your approach doesn't work. Please do it yourself, cause I can't fully understand editor event architecture.

TombEditor/FormMain.cs
Lwmte: ~~Please do it yourself, cause I can't fully understand editor event architecture.~~

Done

TombEditor/Controls/PanelRendering3D.cs
Lwmte: Please propose a way how can you set text position with a function. What is inconsistent? Setting X and Y other than 0?

TombEditor/Controls/PanelRendering3D.cs
Lwmte: How can I use editor.SelectedRoom.WorldPos instead of ```deltaRoomPos * 1024```? Please explain.

TombEditor/TombEditor.csproj
Lwmte: I have replaced API code pack with file save dialog hack described here: https://stackoverflow.com/questions/7330111/select-folder-path-with-savefiledialog/23024435#23024435

Lwmte: Now replaced with proper code borrowed from there under MIT license:
https://www.medo64.com/2011/12/openfolderdialog/

TombEditor/EditorActions.cs
Lwmte: The reason why Banderi used LevelFileNameChange is because he needed to update filename in title bar and remore * sign in case changes were saved. I now introduced new event for this and modified SaveLevel function a bit, so it doesn't modify actual level path if there was an error saving file. This is more consistent, because (as example) if user tried to save prj2 file onto CD-ROM drive and failed, there is no point keeping level file path which leads to read-only drive in level settings.

TombEditor/Controls/PanelRendering3D.cs
ActuallyaDeviloper: There are a ton of "Vector3.Project" in "PanelRendering3D.cs" which is all used for the same thing but with different coordinates now. This is because some instances were adjusted by Banderi, while some some others weren't which I think is unintentional.

My suggestion was to introduce a helper function simialar to the following:
```
void DrawDebugString(string message, Matrix transformation)
{
    Vector3 screenPos = Vector3.Project(new Vector(), 0, 0, ClientSize.Width, ClientSize.Height,
        _device.Viewport.MinDepth, _device.Viewport.MaxDepth, transformation);
    screenPos  += new Vector3(-0.02f, 0.04f); // Offset text a little bit
    _debug.AddString(message, screenPos);
}
```

TombEditor/Controls/Panel2DMap.cs
ActuallyaDeviloper: Because this isn't what's going on here. OnMouseDoubleClick is invoked when the mouse button is _depressed_. Movement of a probe does only make sense while the mouse button is pressed.

Further information: https://msdn.microsoft.com/en-us/library/system.windows.forms.control.mousedoubleclick(v=vs.110).aspx

TombEditor/Controls/Panel2DMap.cs
ActuallyaDeviloper: I think Banderi already addressed this issue, at least line 152 of the current revision no longer sets `_currentlyEditedDepthProbeIndex`.

TombEditor/EditorActions.cs
ActuallyaDeviloper: The name in the titel bar is already handled above already using `LevelFileNameChange()`.

TombEditor/Controls/Panel2DMap.cs
Lwmte: But there is no such code for OnMouseDoubleClick... probably me or you confused this code part with other similar code part (currently line 164). For OnMouseDoubleClick, code looks like this:

                            // Add depth probe under mouse pointer
                            currentProbeIndex = _depthBar.DepthProbes.Count;
                            _depthBar.DepthProbes.Add(new DepthBar.DepthProbe(_depthBar));
                            _depthBar.DepthProbes[currentProbeIndex.Value].Position = clickPos;

Probably GH messed something up, there were a lot of your comments without outdated marking for which fixes were already implemented by Banderi.

Lwmte: Btw, can we simplify code like this?


                            // Add depth probe under mouse pointer
                            _depthBar.DepthProbes.Add(new DepthBar.DepthProbe(_depthBar));
                            _depthBar.DepthProbes.Last().Position = clickPos;

ActuallyaDeviloper: You are right, I was looking at a deprecated code snipet.
I am indeed having some trouble with GitHub here. In particular how it tracks comments (or more like doesn't) on changed code.
Oops! Sorry for wasting your time.

Yes, I have an even more terse suggestion for simplification though:
`_depthBar.DepthProbes.Add(new DepthBar.DepthProbe(_depthBar) { Position = clickPos });`

Lwmte: Done ;)

TombEditor/EditorActions.cs
```diff
@@ -26,14 +26,15 @@ public static class EditorActions
...
+            if (!_editor.HasUnsavedChanges)
                 return true;
 
             switch (DarkMessageBox.Show(owner,
                 "Your unsaved changes will be lost. Do you want to save?",
                 description,
                 MessageBoxButtons.YesNoCancel,
-                MessageBoxIcon.Question))
+                MessageBoxIcon.Question,
+                MessageBoxDefaultButton.Button2))
```
Lwmte: I don't think it's a good idea to make "No" button default. User may accidentally push space or enter and editor will silently exit without saving changes.

By the way, maybe you know why "Yes" button is colored differently from "No" and "Cancel"?

TombEditor/Controls/PanelRendering3D.cs
Lwmte: Added your method, thanks! 👍

TombEditor/Controls/OpenFolderDialog.cs
```diff
@@ -0,0 +1,230 @@
...
+    public string DefaultFolder { get; set; }
+
+    /// <summary>
+    /// Gets selected folder.
+    /// </summary>
+    public string Folder { get; private set; }
+
+
+    internal DialogResult ShowDialog(IWin32Window owner) {
+        if (Environment.OSVersion.Version.Major >= 6) {
```
ActuallyaDeviloper, line 34: I think we should add a check here if it the OS is windows.
You can check TombLauncher.cs for `IsWindows`.

TombEditor/Controls/PanelRendering3D.cs
```diff
@@ -220,6 +224,17 @@ private void EditorEventRaised(IEditorEvent obj)
             if (obj is Editor.ConfigurationChangedEvent)
                 Camera.FieldOfView = ((Editor.ConfigurationChangedEvent)obj).Current.Rendering3D_FieldOfView * (float)(Math.PI / 180);
 
+            // Update camera position
+            if (obj is Editor.RoomGeometryChangedEvent)
+            {
+                if(_editor.Mode == EditorMode.Map2D && Camera != null)
+                {
+                    var deltaRoomPos = _editor.SelectedRoom.Position - _currentRoomLastPos;
+                    Camera.MoveCameraLinear(deltaRoomPos * 1024);
```
ActuallyaDeviloper, line 233: I still think this is wrong on the Y axis because the factor on Y is only 256 (clicks instead sectors!).
We should use `WorldPos` of the room instead instead multiplying by 1024.

PR #188, opened 2017-11-10
Add drag geometry tool

Lwmte:
Adds VERY handy drag tool, which basically replaces 90% of QAWSEDRF functionality.

I am creating pull request for this particular tool because it's not clear how we should deal with diagonal steps heights yet (because currently they use separate button groups to raise/lower them). Possibly we should think of another way instead of YHUJ.

Also, this tool modifies arrow rotation method, now arrows are rotated on mouse up event instead of mouse down event. This is needed to "pack" selection, arrow rotation and drag functionality into same tool (we will need this behaviour for future group transform tools as well).


TombEditor/Geometry/Block.cs
```diff
@@ -488,8 +685,8 @@ public static bool IsQuad(int hXnZp, int hXpZp, int hXpZn, int hXnZn)
 
         public bool FloorIsQuad => IsQuad(QAFaces[FaceXnZp], QAFaces[FaceXpZp], QAFaces[FaceXpZn], QAFaces[FaceXnZn]);
         public bool CeilingIsQuad => IsQuad(WSFaces[FaceXnZp], WSFaces[FaceXpZp], WSFaces[FaceXpZn], WSFaces[FaceXnZn]);
-        public bool FloorHasSlope => FloorDiagonalSplit == DiagonalSplit.None && FloorMax - FloorMin > 2;
-        public bool CeilingHasSlope => CeilingDiagonalSplit == DiagonalSplit.None && CeilingMax - CeilingMin > 2;
+        public bool FloorHasSlope => FloorMax - FloorMin > 2;
+        public bool CeilingHasSlope => CeilingMax - CeilingMin > 2;
```
ActuallyaDeviloper: This change puzzles me a bit. Why are you removing the test for `DiagonalSplit`?

Also would rather suggest the opposite: Adding `FloorDiagonalSplit == DiagonalSplit.None` to the properties `FloorIsQuad` and analogously for `CeilingIsQuad` because otherwise it's not really a quad, right?

TombEditor/Geometry/Block.cs
```diff
@@ -286,57 +441,99 @@ public short GetTriangleMinimumFloorPoint(int triangle)
             }
         }
 
-        public Direction[] GetTriangleSlopeDirections()
+        public Direction[] GetFloorTriangleSlopeDirections()
         {
-            var normals = GetTriangleNormals();
+            var normals = GetFloorTriangleNormals();
```
ActuallyaDeviloper, line 452: I think it is easier to reason about code if we use coordinate descriptions like `SlopedInXp`, `SlopedInZp`, ... and so on instead `North`, `South`, ... because we don't have defined clearly which axes correspond do which cardinal directions and it's a burden to visualize and remember.

Lwmte: The problem is, Direction enum is used not only for slope directions, but also for GetFaces function and maybe somewhere else... that's why I renamed enum from SlideDirection to Direction in the first place. However, it's not a problem renaming it to Direction.Xp, Direction.Xn, etc. I will do that :)

TombEditor/Geometry/Block.cs
Lwmte: Oh, I removed it when I figured out that you actually can have slope on "primary" diagonal step. This was needed for IsIllegalSlope function, because it falsely identified diagonal steps with slopes as no-slope blocks.

Lwmte: Yeah, you're right, FloorIsQuad may return wrong results when primary step has slope with height equal to step's own height... probably that's why IsIllegalSlope was failing for diagonal steps in some cases! I will check that.

PR #585, opened 2022-08-03
621 sounds check all sfx in wadtool

Stranger1992:
Some last minute checks and additions before the TEN Beta release.

Not much has changed but I have added the TR1-TR3 cutscene objects as named slots and no longer "Moveable xxx" .


Stranger1992 (commented):
This is the most important commit to the TEN Beta version of TE as it updates the very outdated version of the soundmap to the latest one.

PR #586, opened 2022-08-04
Trigger manager

Lwmte:
New UI and workflow for trigger volumes in TEN.


Adngel (commented):
This afternoon I already checked many of these codes, I did not see any inconvenient then and neither now. Many of them are the visual trigger window. The rest of the workflow, continue the way I also expected.

I got confussed some moments like the reason to use many small chunks instead of few big chunks, but reading deeply I think it was to use functions orientated to the chunk management, looks like more modern way than using the raw UTF8 instructions. So I think is fine too.

On tests, I did not got any error (apart of my own human errors), still planning to use this feature a lot so if I spot something I'll communicate.

On impression, I think the window is organized and clear, my fears are more related to how it will feel during an actual level development, is possible users may have to create tons of events (even duplicate them to get an individual event-volume behaviour too), that could make the list very long and I'm not sure if it will be confortable navigate throug it. Some sorting methods  for the event list could help to aliviate the potential fuss. But is an extra feature that could be implemented in the future, so I haven't got inconvenient in aprove the branch in this current state.

PR #605, opened 2022-09-06
Missing Object ID's 

Stranger1992:
Addition of various Object ID's from Legacy games that are missing.


Lwmte:
As stated in https://github.com/MontyTRC89/TombEngine/pull/724

PR #618, opened 2022-10-11
Ledge Jumps

goldlife2904:
Ledge Jumps Animations


Lwmte:
Asset changes should be implemented alongside with TEN code changes.

PR #699, opened 2023-05-19
added node  'If Lara is on the vehicle of type...'

davidmarr:
added node 'If Lara is on the vehicle of type...' to check which vehicle Lara is driving

### DIFF
--- a/TombLib/TombLib/Catalogs/TEN Node Catalogs/Lara.lua
+++ b/TombLib/TombLib/Catalogs/TEN Node Catalogs/Lara.lua
@@ -6,7 +6,7 @@
 -- !Arguments "Numerical, 20, Air value, [ 0 | 1800 ]"
 
 LevelFuncs.Engine.Node.TestLaraAir = function(operator, value)
-    return LevelFuncs.Engine.Node.CompareValue(TEN.Objects.Lara:GetAir(), value, operator)
+	return LevelFuncs.Engine.Node.CompareValue(TEN.Objects.Lara:GetAir(), value, operator)
 end
 
 -- !Name "Modify air value"
@@ -16,7 +16,6 @@ end
 -- !Arguments "Numerical, [ -1000 | 1000 | 0 | 1 | 5 ], 15, Air value to define"
 
 LevelFuncs.Engine.Node.ModifyLaraAir = function(operation, value)
-
 	if (operation == 0) then
 		TEN.Objects.Lara:SetAir(TEN.Objects.Lara:GetAir() + value)
 	else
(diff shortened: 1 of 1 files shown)


TombLib/TombLib/Catalogs/TEN Node Catalogs/Lara.lua
```diff
@@ -6,7 +6,7 @@
 -- !Arguments "Numerical, 20, Air value, [ 0 | 1800 ]"
 
 LevelFuncs.Engine.Node.TestLaraAir = function(operator, value)
-    return LevelFuncs.Engine.Node.CompareValue(TEN.Objects.Lara:GetAir(), value, operator)
+	return LevelFuncs.Engine.Node.CompareValue(TEN.Objects.Lara:GetAir(), value, operator)
```
Jakub768, line 9: why are there intends like these everywhere? Is it github being github?

Kubsy (approved):

TombLib/TombLib/Catalogs/TEN Node Catalogs/Lara.lua
davidmarr: VS Code automatically formats all rows by doing indentation

Lwmte: Can you rollback all formatting and keep it the same as before?

TombLib/TombLib/Catalogs/TEN Node Catalogs/Lara.lua
```diff
@@ -163,7 +159,30 @@ end
...
+end
+
+-- !Name "if Lara is on current vehicle..."
+-- !Section "Lara state"
+-- !Conditional "True"
+-- !Arguments "Enumeration, [ Snowmobile | Quad | Speedboat | Kayak | Upv | Minecart | Jeep | Motorbike | Rubber Boat ], 26, type of vehicle"
+-- !Description "Checks if Lara is currently mounting a vehicle and which vehicle it is."
+
+LevelFuncs.Engine.Node.TestLaraVehicleAndType = function(type)
+	local vehicleName = (type == 0) and 'SNOWMOBILE'
```
Lwmte: You need to use `Objects.ObjID` constants instead of hardcoded strings here. Otherwise, if object names will be changed under the hood, this node function will break.

TombLib/TombLib/Catalogs/TEN Node Catalogs/Lara.lua
```diff
@@ -163,7 +159,30 @@ end
 -- !Description "Checks if Lara is currently mounting a vehicle."
 
 LevelFuncs.Engine.Node.TestLaraVehicle = function()
-    return TEN.Objects.Lara:GetVehicle() ~= nil
+	return TEN.Objects.Lara:GetVehicle() ~= nil
+end
+
+-- !Name "if Lara is on current vehicle..."
```
Lwmte: Please rename it to something more descriptive. For example, "If current Lara vehicle is...". Also "If" should start from capital I.

Lwmte (approved):

Issue #706, opened 2023-07-02
TE: Object textures get corrupted when too many object textures are displayed at once


Raildex:
A level by illyaine has many wad objects with big textures.
Displaying all Rooms in TE results in corrupted textures and slowdowns (because it tries to allocate more space).

Jakub768:
tested Raildex's PR and comes up with this error when trying to "draw all rooms" 

```
Application does not support just-in-time (JIT)
debugging. See the end of this message for details.

************** Exception Text **************
System.NullReferenceException: Object reference not set to an instance of an object.
   at TombEditor.Controls.Panel3D.Panel3D.DrawMoveables(List`1 moveablesToDraw, List`1 textToDraw, Boolean disableSelection)
   at TombEditor.Controls.Panel3D.Panel3D.DrawScene()
   at TombEditor.Controls.Panel3D.Panel3D.OnDraw()
   at TombLib.Controls.RenderingPanel.OnPaint(PaintEventArgs e)
   at System.Windows.Forms.Control.PaintWithErrorHandling(PaintEventArgs e, Int16 layer)
   at System.Windows.Forms.Control.WmPaint(Message& m)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ScrollableControl.WndProc(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, WM msg, IntPtr wparam, IntPtr lparam)
```

this does not happen in master

Raildex:
I have reduced the texture slice of the array slices to 1024x1024.
The wad you sent contains textures larger than that.

I would prefer to not restore it to 4096x4096, since that would (on average) waste more memory than needed.

Why does Wad2 allow arbitrary sized textures without a limit? A texture bigger than 4096x4096 would result in the same exception.

PR #707, opened 2023-07-02
Increased WadRenderer Texture Atlas Size

Raildex:
Temporary fix for #706

Could possibly crash on platforms that do not have enough space for uncompressed RGBA8 textures of size 8192x8192 (256 MB)


Lwmte (changes requested):
As stated in Discord, simply increasing atlas size may break compatibility with older systems. You need to recognize available VRAM and only increase atlas size if there is enough memory.

PR #730, opened 2023-11-05
Updated `Draw text...` and `Draw display sprite` nodes

davidmarr:
Updated descriptions with appropriate values
"Draw text..." nodes modified. Added enums for effects


Lwmte:
I think you forgot to update `_System.lua` file, therefore effects, such as blinking, does not work.

PR #733, opened 2023-11-13
Event set editor enhancements

Lwmte:
- Add ability to sort and move event sets in a list (thanks to @Nickelony).
- Add ability to resize event set list.
- Add new global event types (on save, on load, on loop, on start, on end).
- Add nodes to enable and disable events.


davidmarr:
I have done some testing and some minor bugs are present:
- [ ] When I change the name of an event set that is not the first in the list, the first in the list is selected, after the name change.
- [ ] When I save and reload a prj, only the first item in the list of "global event sets" is saved, the others are lost

PR #743, opened 2023-12-05
Splitted the catalog into separate files

Raildex:
Each file is a conformant XML document.

At startup, the Catalogs are combined.
Except for a new function and path adjustments, nothing changed on TE/WadTool side


SoundTool/Program.cs
```diff
@@ -71,12 +71,12 @@ public static void Main(string[] args)
                 Application.SetCompatibleTextRenderingDefault(false);
                 Application.AddMessageFilter(new ControlScrollFilter());
 
-                if (!File.Exists(Application.StartupPath + "\\Catalogs\\TrCatalog.xml"))
+                if (!Directory.Exists(Path.Combine(Application.StartupPath, "Catalogs")))
                 {
                     MessageBox.Show("TrCatalog.xml is missing.\nMake sure you have TrCatalog.xml in /Catalogs/ subfolder.");
```
Lwmte: Can you also change warning to reflect catalog changes?

TombLib/TombLib/Wad/Catalog/TrCatalog.cs
```diff
@@ -455,10 +456,38 @@ public static string GetVersionString(TRVersion.Game version)
...
-            XmlDocument document = new XmlDocument();
-            document.Load(fileName);
+			XmlDocument combinedCatalog = new XmlDocument();
+            combinedCatalog.AppendChild(combinedCatalog.CreateXmlDeclaration("1.0", "UTF-8",null));
+            combinedCatalog.AppendChild(combinedCatalog.CreateElement("trcatalog"));
+			foreach (var gameString in TRVersion.NativeVersions.Select(v => v.Native().ToString()))
+			{
+				var gameCatalogFolder = Path.Combine(rootFolder, gameString);
+				if (!Directory.Exists(gameCatalogFolder))
+					throw new InvalidDataException($"The folder {gameCatalogFolder} does not exist for the Game {gameString}");
```
Lwmte: I am not sure it should throw exception here. User may intentionally delete some folders, e.g. TEN folder (like they do with .dll files in TRNG). For this case, warning should be more appropriate.

Lwmte (changes requested):
All filenames should be in PascalCase.

PR #744, opened 2023-12-06
add FormCamera and FormBumpMaps in XAML

davidmarr:
add the forms **_FormCamera_** and **_FormBumpMaps_** in XAML format


TombEditor.WPF/ToolWindows/RoomOptions.xaml
```diff
@@ -70,7 +70,8 @@
 					<ComboBox
 						x:Name="comboFlipMap"
 						Grid.Row="1"
-						Grid.Column="1">
+						Grid.Column="1"
+                        Width="58">
```
Nickelony: Is there any point in hardcoding the width of this when the `ColumnDefinition` above is set to `Auto`. It will get the minimum required width anyway, right?

TombEditor.WPF/Forms/FormBumpMaps.xaml
```diff
@@ -0,0 +1,37 @@
...
+            <RowDefinition/>
+            <RowDefinition Height="Auto"/>
+            <RowDefinition Height="Auto"/>
+        </sg:SpacedGrid.RowDefinitions>
+        <darkCtrl:SearchableComboBox/>
+        <WindowsFormsHost x:Name="ViewportHost" Grid.Row="1" />
+        <DockPanel Grid.Row="2">
+            <CheckBox Uid="customFile" Content="Use custom file:" Margin="0,0,4,0"/>
+            <Label Content="path to custom file" Opacity="0.4"/>
+        </DockPanel>
```
Nickelony: What is this `DockPanel` used for? To stack the items next to each other? I'd personally use a `StackPanel` for this, but I guess this is a matter of preference, so you don't really have to change this, I'm just wondering. :)

Nickelony (changes requested):

TombEditor.WPF/Forms/FormBumpMaps.xaml
davidmarr: To get the same result with `StackPanel` you have to insert the `Orientation` attribute, it is extra code.

```xaml
<StackPanel Grid.Row="2" Orientation="Horizontal">
    <CheckBox Uid="customFile" Content="Use custom file:" Margin="0,0,4,0"/>
    <Label Content="path to custom file" Opacity="0.4"/>
</StackPanel>
```

Are there performance differences between StackPanel and DockPanel?

TombEditor.WPF/ToolWindows/RoomOptions.xaml
davidmarr: I made that change to RoomOptions.xaml because the space was too tight for the content

|                                                   before                                                  | after                                                                                                     |
|:---------------------------------------------------------------------------------------------------------:|-----------------------------------------------------------------------------------------------------------|
| image | image |

TombEditor.WPF/Forms/FormBumpMaps.xaml
Nickelony: > To get the same result with `StackPanel` you have to insert the `Orientation` attribute, it is extra code.
> 
> ```
> <StackPanel Grid.Row="2" Orientation="Horizontal">
>     <CheckBox Uid="customFile" Content="Use custom file:" Margin="0,0,4,0"/>
>     <Label Content="path to custom file" Opacity="0.4"/>
> </StackPanel>
> ```
> 
> Are there performance differences between StackPanel and DockPanel?

To be honest, I would still use `StackPanel`, even if you must set the `Orientation` because when we eventually port the app to Avalonia, we will replace all StackPanel **item**'s Margins with the dedicated StackPanel `Spacing` property that Avalonia provides for the control. Avalonia uses this property to evenly space out items in a StackPanel, which is not possible in Avalonia's DockPanel control.

PR #745, opened 2023-12-06
Splitted the Catalogs into different files

Raildex:
Internal catalog is merged upon application load
Removed old TrCatalog


Stranger1992 (approved):
✅ Wadtool launches correctly
✅ Tomb Editor launches correctly.
✅ TIDE launches correctly.
✅ Soundtool launches correctly.
✅ Converting ` .PHD` , `.TR2`,`.TR4`,`.TRC` to TombEngine auto-mapping (sounds, slots, state and anim names)
✅ Check if an asset still can be updated with the human-readable state / anim names if found inside the respective catalogue.

Approved from me! Well done. However: one small thing that has annoyed me for a long time is the "t5m" flag that is obviously reference to when TombEngine was called Tomb5Main (whatever that actually means). I think now is a good opportunity to replace that to "TEN" . I will raise a PR for this separately based on this branch.

Lwmte (approved):

Issue #464, opened 2020-06-14
TE: Internally use smaller click height


Lwmte:
In TR5Main chat it was decided that new standard for "click height" should be 64 units, i.e. 4 times smaller than classic one.

Because of this, all TE internal structures and functions must be refactored and internally use smaller click heights. When T5M is set as target, editor should allow more precise editing of corner heights with increment/decrement of 64. With classic targets, it should internally multiply increment by 4. Also, classic compilers should internally divide and round click height to 4 when compiling collisional floordata. In this case, classic targets will be implicitly compatible with T5M projects in the same way triangulated blocks are now compatible with TR2 (visually geometry remains triangulated, but collisional data falls back to legacy quads).

PR #768, opened 2024-01-17
Adjustable step height

Nickelony:
#464

This is achieved through removing the `Level.HeightUnit` const completely and storing geometry height in raw world units as an `int`.
Older prj2 files will have their Y values multiplied by 256 to match original world size.
Step height can be switched around using `[` `]` keys.

ToDo:
- [x] Fix DepthBar
- [x] Add indicator (to indicate whether Precise Mode is ON or OFF)
- [x] Test Room import / export
- [x] Test pathfinding of enemies who can jump / use monkey bars
- [x] Test how precise portals act in older engines
- [x] Adjust rounding for legacy engines
- [x] Find perfect balance for RayTracing rounding
- [x] Test saving prj2 files with complex sectors
- [x] Read through all changes again and make sure everything looks all right

Known issues:
- [x] Slope detection is incorrect
- [x] DepthBar acts weird
- [x] Portal digger uses 1 world unit precision
- [x] Lighting looks different than originally
- [x] Pathfinding of enemies who can jump / use monkey bars is messed up, completely in TR4 and partially in TEN.


Nickelony:
I think it's done!
The main thing I'm worrying about is pathfinding of enemies who can jump / use monkey bars.
This needs to be tested both in TEN and TR4 / TRNG.
All other engines need to be tested; this includes:
- Lighting (whether it looks the same as before)
- Step collisions (whether full 256 unit steps act the same as before)
- Portal behaviour

TombLib/TombLib/LevelData/RoomGeometry.cs
```diff
@@ -1988,10 +1988,11 @@ private static bool RayTraceCheckFloorCeiling(Room room, int x, int y, int z, in
                 return false;
 
             Block block = room.Blocks[currentX, currentZ];
-            int floorMin = block.Floor.Min;
-            int ceilingMax = block.Ceiling.Max;
+            int floorMin = Clicks.FromWorld(block.Floor.Min);
+            int ceilingMax = Clicks.FromWorld(block.Ceiling.Max);
+            int yClicks = Clicks.FromWorld(y);
```
Nickelony, line 1993: This might actually backfire. I will look into this later.

TombLib/TombLib/LevelData/Compilers/Rooms.cs
```diff
@@ -1315,35 +1315,29 @@ private void ConvertSectors(Room room, tr_room newRoom)
                     if (x == 0 || z == 0 || x == room.NumXSectors - 1 || z == room.NumZSectors - 1 ||
                         block.Type == BlockType.BorderWall || block.Type == BlockType.Wall)
                     {
-                        sector.Floor = (sbyte)(-room.Position.Y - block.Floor.Max);
-                        sector.Ceiling = (sbyte)(-room.Position.Y - block.Ceiling.Min);
+                        sector.Floor = (sbyte)(-Clicks.FromWorld(room.Position.Y) - Clicks.FromWorld(block.Floor.Max));
```
Nickelony, line 1318: These calculations can be further optimized. Will do later.

Nickelony:
It's done!
In the future we may allow the user to choose which type of rounding (Default, ToFloor, ToCeiling or Integer) should be used for RayTracing since each one of them makes the lighting look quite different, but for now, this is as close to original as possible.

TombEditor/Editor.cs
```diff
@@ -5,6 +5,7 @@
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Windows.Forms;
```
Lwmte: Do we need this include here?

TombEditor/Controls/DepthBar.cs
```diff
@@ -211,7 +211,7 @@ public bool MouseDown(MouseEventArgs e, Size parentControlSize, Vector2 clickPos
                                             if (mouseDepth <= roomSequences[i][j].MaxDepth && mouseDepth >= roomSequences[i][j].MinDepth)
                                             {
                                                 _roomMouseClicked = roomSequences[i][j].Room;
-                                                _roomMouseOffset = mouseDepth - _roomMouseClicked.Position.Y;
+                                                _roomMouseOffset = mouseDepth - Clicks.FromWorld(_roomMouseClicked.Position.Y);
```
Lwmte, line 214: I am not fully understanding why `Clicks.From/ToWorld()` methods are used for room coordinates. Was room vertical coordinate format changed as well, together with variable click height?

TombEditor/Controls/Panel3D/Panel3DText.cs
```diff
@@ -53,7 +53,7 @@ private static string GetObjectRotationString(Room room, PositionBasedObjectInst
         private static string GetObjectPositionString(Room room, PositionBasedObjectInstance instance)
         {
             // Get the distance between point and floor in units
-            int height = (int)(instance.Position.Y - GetFloorHeight(room, instance.Position)) / (int)Level.HeightUnit;
+            float height = (instance.Position.Y - GetFloorHeight(room, instance.Position)) / Level.FullClickHeight;
```
Lwmte, line 56: Why this variable was converted to float, if in all other parts coordinates are seemingly converted to int?

TombLib/TombLib/LevelData/Compilers/TombEngine/PathfindingDecompiled.cs
```diff
@@ -32,10 +32,10 @@ public sealed partial class LevelCompilerTombEngine
         private bool dec_flipped;
         private bool dec_jump;
         private Room dec_currentRoom;
-        private short dec_q0 = -1;
-        private short dec_q1 = -1;
-        private short dec_q2 = -1;
-        private short dec_q3 = -1;
+        private int dec_q0 = Clicks.ToWorld(-1);
```
Lwmte, line 35: I don't understand this. Why these 4 variables were converted to full world unit scale, if all other underlying code remains the same?

TombLib/TombLib/LevelData/Compilers/TombEngine/PathfindingDecompiled.cs
```diff
@@ -830,10 +830,10 @@ private int Dec_GetBoxFloorHeight(int x, int z, out bool slope)
             dec_q2 = block.Floor.XpZn;
             dec_q3 = block.Floor.XnZn;
 
-            int slope1 = Math.Abs(dec_q0 - dec_q1) >= 3 ? 1 : 0;
-            int slope2 = Math.Abs(dec_q1 - dec_q2) >= 3 ? 1 : 0;
-            int slope3 = Math.Abs(dec_q2 - dec_q3) >= 3 ? 1 : 0;
-            int slope4 = Math.Abs(dec_q3 - dec_q0) >= 3 ? 1 : 0;
+            int slope1 = Math.Abs(dec_q0 - dec_q1) >= Clicks.ToWorld(3) ? 1 : 0;
```
Lwmte, line 833: Same as above

TombLib/TombLib/LevelData/Compilers/TombEngine/PathfindingDecompiled.cs
```diff
@@ -866,7 +866,7 @@ private int Dec_GetBoxFloorHeight(int x, int z, out bool slope)
             int floorHeight = meanFloorCornerHeight + room.Position.Y;
             int ceiling = block.Ceiling.Max + room.Position.Y;
 
-            if (dec_water && room.Properties.Type == RoomType.Water && ceiling - meanFloorCornerHeight <= 1 && block.CeilingPortal != null)
+            if (dec_water && room.Properties.Type == RoomType.Water && ceiling - meanFloorCornerHeight <= Clicks.ToWorld(1) && block.CeilingPortal != null)
```
Lwmte, line 869: Same as above

TombEditor/Editor.cs
Nickelony: No, not anymore, good catch!

TombEditor/Controls/DepthBar.cs
Nickelony: FromWorld -> value / 256 (e.g. 512 / 256 = 2 clicks)
ToWorld -> value * 256 (e.g 2 clicks * 256 = 512 world units)

We store height in world units now exclusively, but some calculations rely on Click height anyway and I preferred not to fiddle with them unless absolutely necessary.

TombEditor/Controls/Panel3D/Panel3DText.cs
Nickelony: To display that the object is 0.25 clicks above the ground, otherwise it would only show 0.

TombLib/TombLib/LevelData/Compilers/TombEngine/PathfindingDecompiled.cs
Nickelony: Because `dec_q0` now stores world units instead of clicks (at least for TEN).
Honestly, Pathfinding code is a mystery to me, because it seems that it doesn't actually go anywhere most of the time.
Also keeping it in Clicks made the AI go wild originally, but maybe it was something else. I wouldn't risk changing it back though.

Lwmte: You mean that AI now behaves better than before this change? If so, it may be another incorrect decompilation which got fixed only now.

Lwmte (approved):

PR #801, opened 2024-04-10
Add default texture functionality

Lwmte:


Stranger1992:
I cant comment on the internal code here, but functionally works really nicely!

PR #815, opened 2024-05-01
Fix (and de-harcode) texture sounds "limit" when loading a PRJ2

### LINKED ISSUE #814: Texture sounds Custom3 to Custom8 revert to Custom2 after reloading
**Steps to reproduce:**
1. Open a project with textures.
2. Open the Texture sounds window.
3. Assign multiple footstep sounds to multiple textures, at least covering sounds Custom3 to Custom8.
   Example:
   image

4. Save the project. You can build and play, it works as intended.
5. Now close the project (either open a different project, or close TE).
6. Load the project again in TE.
   At this stage, all texture sounds are restored EXCEPT Custom3 to Custom8: they're all reverted to Custom2.
   image

It is not a problem of the **Texture sounds** window itself, because if you build the level right after reloading the project, without even re-opening the Texture sounds window in between, you can still notice in-game that the info has been lost.

So it is either a problem with saving or loading a project.

**Note**: before TombEngine, TE only listed 16 texture sounds entries. All of them had names, except the last 2: at the time they were named Unknown14 and Unknown15. They actually correspond to current Custom1 and Custom2.
The list was extended for TombEngine specifically, so something was probably forgotten in the process.

Joey79100:
This fixes #814 : loading a PRJ2 was discarding any texture sound ID above 15.

This PR "raises the limit", and also de-hardcodes it by referring to the enum's length instead.
However I'm not sure if it's ok to de-hardcode: I don't know anything about chunks, so I'm not sure why this check was added in the beginning, maybe there's something I'm not seeing.

Just in case de-hardcoding is a bad idea, I first commited the "limit raise" from 15 to 22, then I commited the de-harcoding (so you can still cherry-pick the 1st one).

### DIFF
--- a/TombLib/TombLib/LevelData/IO/Prj2Loader.cs
+++ b/TombLib/TombLib/LevelData/IO/Prj2Loader.cs
@@ -405,8 +405,8 @@ namespace TombLib.LevelData.IO
                                     for (int x = 0; x < levelTexture.FootStepSoundWidth; ++x)
                                     {
                                         byte textureSoundByte = chunkIO.Raw.ReadByte();
-                                        if (textureSoundByte > 15)
-                                            textureSoundByte = 15;
+                                        if (textureSoundByte > 21)
+                                            textureSoundByte = 21;
                                         levelTexture.SetFootStepSound(x, y, (TextureFootStep.Type)textureSoundByte);
                                     }
                             }
--- a/TombLib/TombLib/LevelData/IO/Prj2Loader.cs
+++ b/TombLib/TombLib/LevelData/IO/Prj2Loader.cs
@@ -405,8 +405,8 @@ namespace TombLib.LevelData.IO
                                     for (int x = 0; x < levelTexture.FootStepSoundWidth; ++x)
                                     {
                                         byte textureSoundByte = chunkIO.Raw.ReadByte();
-                                        if (textureSoundByte > 21)
-                                            textureSoundByte = 21;
+                                        if (textureSoundByte > (byte)Enum.GetNames(typeof(TextureFootStep.Type)).Length)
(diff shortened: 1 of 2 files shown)


TombLib/TombLib/LevelData/IO/Prj2Loader.cs
```diff
@@ -405,8 +405,8 @@ private static bool LoadLevelSettings(ChunkReader chunkIO, ChunkId idOuter, stri
                                     for (int x = 0; x < levelTexture.FootStepSoundWidth; ++x)
                                     {
                                         byte textureSoundByte = chunkIO.Raw.ReadByte();
-                                        if (textureSoundByte > 15)
-                                            textureSoundByte = 15;
+                                        if (textureSoundByte > (byte)Enum.GetNames(typeof(TextureFootStep.Type)).Length)
+                                            textureSoundByte = (byte)Enum.GetNames(typeof(TextureFootStep.Type)).Length;
```
Nickelony: This will fail, as you're assigning a value that's larger than the last enum element.
Additionally, I wouldn't use the `Enum.GetNames()` method as I think it uses Reflection underneath.
Please add a value at the end of the enum called `Count` and use that one instead:
image

Then replace those 2 lines you edited with this:
```cs
byte maxSoundByte = (byte)(TextureFootStep.Type.Count - 1);

if (textureSoundByte > maxSoundByte)
	textureSoundByte = maxSoundByte;
```

JoeyQuint: Thanks, indeed I forgot the - 1.
Just did what you asked, and edited `TextureFootStep.GetNames()` to avoid having `Count` appear in the dropdown.

PR #818, opened 2024-05-10
Keep RoomGeometry allocated

Raildex:
Keeps RoomGeometry allocated for a single room.

Instead of Destroying and Recreating the RoomGeometry, this PR keeps the RoomGeometry created once per room.
This has the advantage that a single geometry edit does not result in a cascade of object destructions and constructions.
This also keeps the capacity of the Lists within RoomGeometry, i.e. the Lists keep their memory instead of constructed with a low capacity and then reallocating often during vertex insertion.

### DIFF
--- a/TombLib/TombLib/LevelData/Room.cs
+++ b/TombLib/TombLib/LevelData/Room.cs
@@ -129,7 +129,7 @@ namespace TombLib.LevelData
         public short AlternateGroup { get; set; } = -1;
 
         // Internal data structures
-        public RoomGeometry RoomGeometry { get; set; }
+        public RoomGeometry RoomGeometry { get;} = new RoomGeometry();
 
         public Room(Level level, int numXSectors, int numZSectors, Vector3 ambientLight, string name = "Unnamed", int ceiling = DefaultHeight)
         {
@@ -193,7 +193,6 @@ namespace TombLib.LevelData
                 }
 
             // Update data structures
-            RoomGeometry = null;
             Blocks = newBlocks;
 
             // Move objects
@@ -322,8 +321,6 @@ namespace TombLib.LevelData
                 result.AlternateRoom = null;
                 result.AlternateGroup = -1;
             }
-
-            result.RoomGeometry = null;
 
             // Copy properties
             result.Properties = Properties.Clone();
@@ -921,7 +918,7 @@ namespace TombLib.LevelData
 
         public void BuildGeometry(bool legacy = false)
         {
-            RoomGeometry = new RoomGeometry(this, legacy);
+            RoomGeometry.Build(this);
         }
 
         public void RebuildLighting(bool highQualityLighting)
@@ -1532,8 +1529,7 @@ namespace TombLib.LevelData
             // Update adjoining rooms
             var roomsToProcess = new List<Room> { this };
             var areaToProcess = new List<RectangleInt2> { area };
-            List<PortalInstance> listOfPortals = Portals.ToList();
-            foreach (var portal in listOfPortals)
+            foreach (var portal in Portals)
             {
                 if (!portal.Area.Intersects(area))
                     continue; // This portal is irrelavant since no changes happend in its area
--- a/TombLib/TombLib/LevelData/Room.cs
+++ b/TombLib/TombLib/LevelData/Room.cs
@@ -917,7 +917,7 @@ namespace TombLib.LevelData
 
         public void BuildGeometry(bool legacy = false)
         {
-            RoomGeometry.Build(this);
+            RoomGeometry.Build(this, legacy);
         }
 
         public void RebuildLighting(bool highQualityLighting)
--- a/TombLib/TombLib/LevelData/Room.cs
+++ b/TombLib/TombLib/LevelData/Room.cs
@@ -313,8 +313,7 @@ namespace TombLib.LevelData
         public Room Clone(Level level, Predicate<ObjectInstance> decideToCopy, bool fullCopy = false)
         {
             // Copy most variables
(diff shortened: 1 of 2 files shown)


TombLib/TombLib/LevelData/RoomGeometry.cs
```diff
@@ -43,10 +43,21 @@ public class RoomGeometry
...
+        }
+        public void Build(Room room, bool legacy = false)
+        {
+            VertexPositions.Clear();
+            VertexEditorUVs.Clear();
+            VertexColors.Clear();
+            TriangleTextureAreas.Clear();
+            TriangleSectorInfo.Clear();
+            SharedVertices.Clear();
+            VertexRangeLookup.Clear();
```
Nickelony, line 53: Could you please benchmark if clearing the collections causes the process to be slower, because that's what I was worrying about when also considering this approach.

Try and create a 100x100 room in TEN and test on that as it takes a very long time to process a room like this.

TombLib/TombLib/LevelData/Room.cs
```diff
@@ -921,7 +917,7 @@ public BlockFaceShape GetFaceShape(int x, int z, BlockFace face)
 
         public void BuildGeometry(bool legacy = false)
```
Nickelony, line 918: You're not using the `legacy` bool anymore. Please remove the argument and make sure that importing legacy .PRJ files still yields correct results, as using current geometry code breaks legacy .PRJs in very rare cases, such as `coastal.prj` having wrong textures in the coast area, near the hidden grenades underwater.

Nickelony (changes requested):

TombLib/TombLib/LevelData/RoomGeometry.cs
Raildex: In the case of lists and dictionaries, clearing does not change the capacity. 
This is better than recreating Lists and Dictionaries with a low capacity and then re-filling them; During the refill the collections would need to realloc and rehash their contents. This is not the case when doing `Clear()`.

TombLib/TombLib/LevelData/Room.cs
Raildex: Can you give an example?

Nickelony: Will do after I finish work in around 2.5h

Raildex:
@Nickelony Is legacy code determined at runtime? if not, just use a compile constant instead of propagating it in the whole call chain

Nickelony:
> @Nickelony Is legacy code determined at runtime? if not, just use a compile constant instead of propagating it in the whole call chain

Yes, it's determined at runtime, specifically during the conversion from legacy .PRJ to .PRJ2.

Here is current geometry code used on coastal.prj:
image

Legacy code result:
image

Current geometry code causes a few textures to be missing as they are evaluated later during the conversion process.

TombLib/TombLib/LevelData/RoomGeometry.cs
```diff
@@ -554,7 +549,7 @@ private enum FaceDirection
             }
         }
 
-        private void AddVerticalFaces(Room room, int x, int z, FaceDirection faceDirection, bool hasFloorPart, bool hasCeilingPart, bool hasMiddlePart)
+        private void AddVerticalFaces(Room room, int x, int z, FaceDirection faceDirection, bool hasFloorPart, bool hasCeilingPart, bool hasMiddlePart, bool useLegacyCode)
```
Raildex, line 552: I'd prefer to not have this parameter carried into the whole call chain

Raildex:
> > @Nickelony Is legacy code determined at runtime? if not, just use a compile constant instead of propagating it in the whole call chain
> 
> Yes, it's determined at runtime, specifically during the conversion from legacy .PRJ to .PRJ2.
> 
> Here is current geometry code used on coastal.prj: image
> 
> Legacy code result: image
> 
> Current geometry code causes a few textures to be missing as they are evaluated later during the conversion process.

Wouldn't that mean that the conversion of the PRJ is incorrect in this case or atleast needs to be adapted into the new geometry creation process?

Nickelony:
> > > @Nickelony Is legacy code determined at runtime? if not, just use a compile constant instead of propagating it in the whole call chain
> > 
> > Yes, it's determined at runtime, specifically during the conversion from legacy .PRJ to .PRJ2.
> > 
> > Here is current geometry code used on coastal.prj: image
> > 
> > Legacy code result: image
> > 
> > Current geometry code causes a few textures to be missing as they are evaluated later during the conversion process.
> 
> Wouldn't that mean that the conversion of the PRJ is incorrect in this case or atleast needs to be adapted into the new geometry creation process?

Yes, but the code is extremely fragile and it's very hard for me to understand it, so this current placeholder needs to stay or we just completely remove the ability to import legacy PRJ files.

TombLib/TombLib/LevelData/RoomGeometry.cs
```diff
@@ -1502,7 +1497,7 @@ private void AddVerticalFaces(Room room, int x, int z, FaceDirection faceDirecti
                 ZB = zB,
             };
 
-            if (_legacy)
+            if (useLegacyCode)
```
Raildex, line 1500: Maybe you can use a delegate instead?
Provide a legacy `AddXYZFace` method and a new `AddXYZFace`.
And set the delegate accordingly?
This will get rid of the if within the method and keeps the two separated.

Nickelony:
The code of this PR is fine, but sadly, there have been some discoveries made in other parts of the TE code that really don't like this implementation and are quite possibly a bad memory leak hazard. Until that has been addressed, this PR needs reverting.

PR #834, opened 2024-05-27
Added TextureArray support for Object Rendering

### LINKED ISSUE #706: TE: Object textures get corrupted when too many object textures are displayed at once

Raildex:
Fixes #706 
This PR changes the Texture handling for **Object** rendering.

Prior, the WadRenderer would utilize a single 4096x4096px texture to store all needed textures during rendering.

This PR changes it to use a Texture2DArray instead. The Texture Packer is now a dynamic list and packed textures are stored in a `VectorInt3` to store the UVs and the Atlas Index.

When the TextureArray is insufficient, a new texture is created with new a new size and old content is copied over.

The changed `WadRenderer` has now a configurable Atlas Size as well as a limit of how big a wad texture can be when packing. When a wad texture is too big, it gets resized to the allowed size.
In the case of TE, the 3D Level view has a Texture Atlas Size of 512x512 and a maximum allowed size of 96x96. I've tested multiple Allowed sizes and came to the conclusion this is a good tradeoff between quality loss and memory footprint.

The TE Item Preview,WT Main Preview have an atlas size of 1024x1024 and allowed size of 128x128.

WT Skeleton and Mesh Editor both have an atlas size of 2048x2048 and maximum allowed size of 512x512, because visual quality is a must here.


Nickelony (approved):
If this is tested and it works, then this is good with me.

Raildex:
I have increased the Texture Size to 4096x4096 to test whether it works correctly.

Slice 0:
grafik
Slice 1:
grafik


In @Jakub768 's Wad, he uses Texture sizes beyond 256x256 with a size of around 1125px, which is why it crashed upon filling the texture.
I would prefer to keep the texture size 1024x1024, so the wastage of memory is kept minimal.

@Jakub768: If I was you, I would optimize your wad so it uses only texture pages of size 256x256 max.

@Lwmte @MontyTRC89 @Nickelony : Wad2 needs a restriction of how big textures can be.
256x256, maybe 512x512 should be the maximum, anything beyond that is absurd.

Lwmte:
We should not impose a limit on a texture page size. 256x256 texture page size was a legacy limit that a lot of users complained about. Also, builder may have an UV mapped model with a texture page as large as possible, but applied onto a high-poly mesh. Imposing an artificial texture page limit would break this advantage.

Raildex:
I've added resizing of the textures with a now adjustable Atlas Size and Maximum Size for the texture.
If a texture is bigger than that, it gets resized.
This now works with even smaller atlas sizes with fairly acceptable quality loss:
grafik
grafik

Jakub768:
> I temporarily increased the Texture Size to 4096x4096 to test whether it works correctly and it does.
> 
> Slice 0: grafik Slice 1: grafik
> 
> In @Jakub768 's Wad, he uses Texture sizes beyond 256x256 with a size of around 1125px, which is why it crashed upon filling the texture. I would prefer to keep the texture size 1024x1024, so the wastage of memory is kept minimal. You can see how much memory is wasted on Slice 1.
> 
> @Jakub768: If I was you, I would optimize your wad so it uses only texture pages of size 256x256 max.
> 
> @Lwmte @MontyTRC89 @Nickelony : Wad2 needs a restriction of how big textures can be. 256x256, maybe 512x512 should be the maximum, anything beyond that is absurd. A theoretical texture of size 4097x4097 would crash as well.

Ow I never knew that my texture atlases are beyond 256x256, probably an issue on my end when retexturing some statics I guess

Jakub768:
I have now tested and as I said to @Raildex and @Lwmte , I can view all of my rooms at once without any serious freezes when

Jakub768 (approved):

Nickelony:
I will do some extra testing next weekend and get this PR finally merged

Nickelony:
All good to go and ready to merge.
@Raildex I just wonder if we should make the lower resolution preview a setting, or is it required? What if someone wants to preview the objects in full res textures?

Raildex:
@Nickelony I suggested such a feature to @Lwmte in the future. But for now I would keep it as is to not clutter the UI even more.

PR #903, opened 2024-10-04
Fix for wadrenderer leaks

Raildex:
- Adds parameter to WadRenderer to load animation data from wad. ItemPreview does not need animations. Animations would occupy hundreds of MB for a large wad.
- Fixes the disposal of the `Mesh<T,U>` class, which did not clean up the vertex and index buffers before
- Fixes the `WadRenderer` texture atlas disposal when the texture atlas needs to be resized
  - Shader parameters were moved to after the model-to-be-rendered was retrieved so an invalid texture is not used anymore. Potentially fixes wrong texture display when loading a new wad or level

### DIFF
--- a/TombEditor/Controls/Panel3D/Panel3DDraw.cs
+++ b/TombEditor/Controls/Panel3D/Panel3DDraw.cs
@@ -1505,13 +1505,7 @@ namespace TombEditor.Controls.Panel3D
         {
             if (moveablesToDraw.Count == 0)
                 return;
-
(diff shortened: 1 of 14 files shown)


Nickelony (approved):

Lwmte:
ItemPreview **needs** animations:

Before your pull request:

Clipboard-2

After your pull request:

Clipboard-1

Your added parameter for `WadRenderer` shouldn't bypass loading of all animation data, but load at least first frame of a first animation.

PR #937, opened 2024-11-07
Fix TR1 pickup definitions

### LINKED ISSUE #804: TR1 Scion holder's Scion is converted to the wrong Scion object number on build
TR1 has multiple Scion objects that seem the same but function differently in game. This causes a crash for custom levels in TR1X at least. Here's the Scion object numbers from TR1 with some comments on their usage:

image

The Scion holder should use Scion object `146`, but Tomb Editor converts it to `142` which is a quest pickup item. For reference, the following fix works which might help solve the issue:

1. Go to `C:\Tomb Editor\Catalogs\Engines\TR1` (or wherever you have TE installed) and open `Moveables.xml` in notepad or equivalent. If you have an older version of TE it's `C:\Tomb Editor\Catalogs\TrCatalog.xml`
2. Look for these lines
```
    <moveable id="145" name="Complete Scion" id2="141" ten="KEY_ITEM1" />
    <moveable id="146" name="Complete Scion" id2="142" ten="KEY_ITEM2" />
```
3. Remove the `id2` attribute altogether on both:
```
    <moveable id="145" name="Complete Scion" ten="KEY_ITEM1" />
    <moveable id="146" name="Complete Scion" ten="KEY_ITEM2" />
```
4. Restart TombEditor and build the level with complete scion in place.
5. Use TRCinematicImporter to get the camera in place too otherwise it crashes when Lara tries to grab it (only needed for Atlantis type, Great Pyramid type doesn't trigger the misc animation).

Example attached showing it can work.

lahm86:
This fixes the name of the lead bar item; defines Pickup 1 and Pickup 2; and, removes the incorrect association of these pickup sprites with the animated scion models.

Tested in TombATI and TR1X.

Resolves #804.

### DIFF
--- a/TombLib/TombLib/Catalogs/Engines/TR1/Moveables.xml
+++ b/TombLib/TombLib/Catalogs/Engines/TR1/Moveables.xml
@@ -108,7 +108,7 @@
     <moveable id="123" name="Slot 2 full" ten="PUZZLE_DONE_2" />
     <moveable id="124" name="Slot 3 full" ten="PUZZLE_DONE_3" />
     <moveable id="125" name="Slot 4 full" ten="PUZZLE_DONE_4" />
-    <moveable id="127" name="Pickup 1" id2="126" ten="PICKUP_ITEM1" />
+    <moveable id="127" name="Lead bar" id2="126" ten="PICKUP_ITEM1" />
     <moveable id="128" name="Midas gold touch" essential="false" />
     <moveable id="133" name="Key 1" id2="129" essential="false" ten="KEY_ITEM1" />
     <moveable id="134" name="Key 2" id2="130" essential="false" ten="KEY_ITEM2" />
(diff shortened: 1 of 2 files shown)


TombLib/TombLib/Catalogs/Engines/TR1/Moveables.xml
```diff
@@ -118,9 +118,11 @@
     <moveable id="138" name="Lock 2" essential="false" ten="KEY_HOLE2" />
     <moveable id="139" name="Lock 3" essential="false" ten="KEY_HOLE3" />
     <moveable id="140" name="Lock 4" essential="false" ten="KEY_HOLE4" />
-    <moveable id="145" name="Complete Scion" id2="141" ten="KEY_ITEM1" />
-    <moveable id="146" name="Complete Scion" id2="142" ten="KEY_ITEM2" />
```
Lwmte, line 122: `id2` field is used in TE to convert visible 3D items in editor back to sprite pickup objects in TR1 and TR2.

lahm86: Yes, but these are wrong. These objects are the animated scions  (Atlantis and Great Pyramid in OG) so they're not pickup objects. To add a normal scion pickup, you would add object 150.
This problem is mentioned in detail in the issue.

Lwmte: Sorry, I remember reading about it somewhere :)

PR #959, opened 2024-11-24
Change shatter sound in WadTool

Lwmte:
Adds controls to static mesh editor to customize shatter attribute and shatter sound. Any static mesh now can be made shatterable and can have unique sound effect ID.

### DIFF
--- a/Installer/Changes.txt
+++ b/Installer/Changes.txt
@@ -17,7 +17,8 @@ Tomb Editor:
  * Added TEN nodes to get and set hit points for a shatterable static mesh.
  
 WadTool:
- * Added "Disable Interpolation" animcommand type.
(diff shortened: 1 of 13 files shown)


Stranger1992 (commented):
# Testing

## Functionality
* Shatter sound changes in Wadtool and is carried over to TombEngine ✅

## Comments

* A reference project is currently required for the shatter sound to be changed.  Can this be changed so you don't need to have this and can still change the shatter sound? On one hand: I understand why it is required so the correct index is used from the soundmap 

* One thing to consider is, in the default naming for the TEN soundmap: ideal shatter sounds have a prefix of "TRX_SMASH_" . Can these indexes be listed first? If not: it doesn't matter but just from the point of view of an end user.

PR #961, opened 2024-11-24
Fixed smaller texture selection sizes not working correctly

Nickelony:


TombLib/TombLib.Forms/Controls/TextureMapBase.cs
```diff
@@ -166,13 +174,13 @@ public SelectionPrecisionType(float precision, bool selectFullTileAutomatically)
             }
         }
 
-        protected virtual SelectionPrecisionType GetSelectionPrecision(bool rectangularSelection)
+        protected virtual SelectionPrecisionType GetSelectionPrecision()
```
Lwmte: Why this argument was removed everywhere? Did you check how it works if you randomly change texture coordinates and make selection non-rectangular (by dragging individual corners of a texture selection)?

Nickelony: I changed the name of the variable so it's easier to understand and forgot to stage the changes. Nice catch!

PR #989, opened 2025-02-25
Alpha tint nodes

TrainWrack:
Adds alpha tint nodes over time


TombLib/TombLib/Catalogs/TEN Node Catalogs/Timespan Actions.lua
```diff
@@ -161,11 +161,29 @@ end
 -- !Description "Gradually change colour of a moveable over specified timespan."
 -- !Arguments "NewLine, Moveables, 70" "Color, 15, Moveable colour"
 -- !Arguments "Numerical, [ 0.1 | 65535 | 2 | 0.1 | 1 ], {1}, 15, Time (in seconds)"
+-- !Arguments "NewLine, Numerical, [ 0 | 255 | 2 | 1 | 1 ], {1}, 15, Alpha"
```
Lwmte: In my opinion, it's better to have separate node for changing transparency, even if it is shared in one field. Builder may want to separately change color and transparency.

TrainWrack: I have added separate nodes already. This one can stay incase they want to change the color and alpha at the same time. The default value can be set to 255 to make sure it always fade color only unless somebody else changes it.

TombLib/TombLib/Catalogs/TEN Node Catalogs/Timespan Actions.lua
```diff
@@ -161,11 +161,29 @@ end
...
 -- !Arguments "Numerical, [ 0.1 | 65535 | 2 | 0.1 | 1 ], {1}, 15, Time (in seconds)"
+-- !Arguments "NewLine, Numerical, [ 0 | 255 | 2 | 1 | 1 ], {255}, 15, Alpha"
 
-LevelFuncs.Engine.Node.ChangeMoveableColourOverTimespan = function(moveableName, newColour, time)
+LevelFuncs.Engine.Node.ChangeMoveableColourOverTimespan = function(moveableName, newColour, time, alpha)
+	newColour.a = alpha
 	do LevelFuncs.Engine.Node.ConstructTimedData(moveableName, false, 3, newColour, time, true) end
 end
 
+-- !Name "Change alpha of a moveable"
```
Lwmte: I suggest to use word "transparency" instead of "alpha". For builders without experience with graphics terminology it may sound as something unknown.

TombLib/TombLib/Catalogs/TEN Node Catalogs/Timespan Actions.lua
Lwmte: I don't agree that we need to modify existing nodes with extra transparency field, especially because alpha value will be last in the UI element group. It will confuse builders. Additionally, I can't imagine a use case when color and transparency should be set at the same time.

PR #996, opened 2025-03-08
Tr5 Laser + TR1 Thor Hammer settings

TrainWrack:
Adds sounds and object slot


Lwmte:
Implemented via separate commit

PR #1020, opened 2025-04-21
Dependency cleanup

Raildex:
This PR removes the Codepages Nuget Package (which made problems before due to different versions in different solution projects) and removes DarkUI from TombLib.

One interface (IHasLuaName) had DarkUI as a dependency so it could open a DarkMessageBox.
The interface operation was renamed from `TrySetLuaName` to `CanSetLuaName` to accomodate the Query-Command pattern (instead of silently failing and setting the name).

UI stuff should only go in the App Projects (TombIDE, WadTool, SoundToo, TombEditor) or the UI library projects (TombLib.Forms, DarKUI) so you do not lock into DarkUI even more.

### DIFF
--- a/DarkUI/DarkUI/DarkUI.csproj
+++ b/DarkUI/DarkUI/DarkUI.csproj
@@ -123,7 +123,4 @@
   <ItemGroup>
     <Content Include="Resources\grip_fill.png" />
   </ItemGroup>
-  <ItemGroup>
-    <PackageReference Include="System.Text.Encoding.CodePages" Version="5.0.0" />
-  </ItemGroup>
 </Project>
--- a/TombEditor/Controls/Panel3D/MouseHandler/Panel3DMouse.cs
+++ b/TombEditor/Controls/Panel3D/MouseHandler/Panel3DMouse.cs
@@ -160,6 +160,7 @@ namespace TombEditor.Controls.Panel3D
         {
             // Check if we are done with all common file tasks
             var filesToProcess = EditorActions.DragDropCommonFiles(e, FindForm());
+
             if (filesToProcess == 0)
                 return;
 
@@ -176,6 +177,7 @@ namespace TombEditor.Controls.Panel3D
                     _editor.SelectedRoom = newSectorPicking.Room;
 
                 var obj = e.Data.GetData(e.Data.GetFormats()[0]) as IWadObject;
+
                 if (obj != null)
                 {
                     PositionBasedObjectInstance instance = null;
@@ -200,11 +202,11 @@ namespace TombEditor.Controls.Panel3D
                     {
                         if (!BaseGeometryImporter.FileExtensions.Matches(file))
                             continue;
+
                         if (!file.CheckAndWarnIfNotANSI(this)) 
                         {
                             DarkMessageBox.Show(FindForm(), "Filename or path is invalid. Please use standard characters.", "Wrong filename", MessageBoxIcon.Error);
                             continue;
-
                         }
 
                         EditorActions.AddAndPlaceImportedGeometry(this, newSectorPicking.Pos, file);
--- a/TombEditor/Controls/Panel3D/MouseHandler/Panel3DMouse.cs
+++ b/TombEditor/Controls/Panel3D/MouseHandler/Panel3DMouse.cs
@@ -1,4 +1,5 @@
 ﻿using DarkUI.Controls;
+using DarkUI.Forms;
 using System;
 using System.Collections.Generic;
 using System.Drawing;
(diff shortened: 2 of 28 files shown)


TombEditor/Controls/Panel3D/MouseHandler/Panel3DMouse.cs
```diff
@@ -200,9 +201,12 @@ private void OnMouseDragAndDrop(DragEventArgs e)
                         if (!BaseGeometryImporter.FileExtensions.Matches(file))
                             continue;
 
-                        if (!file.CheckAndWarnIfNotANSI(this))
+                        if (!file.CheckAndWarnIfNotANSI(this)) {
```
Nickelony: Curly brace should start on next line rather than on same line as the `if ()` statement

TombEditor/Controls/Panel3D/MouseHandler/Panel3DMouse.cs
```diff
@@ -200,9 +201,12 @@ private void OnMouseDragAndDrop(DragEventArgs e)
...
+                            DarkMessageBox.Show(FindForm(), "Filename or path is invalid. Please use standard characters.", "Wrong filename", MessageBoxIcon.Error);
                             continue;
```
Nickelony: Remove unnecessary newline

TombEditor/ObjectClipboardData.cs
```diff
@@ -78,9 +80,17 @@ public ObjectInstance MergeGetSingleObject(Editor editor)
                 {
                     editor.SelectedRoom.AddObject(editor.Level, obj);
                     var luaObj = obj as IHasLuaName;
-                    if (!luaObj.TrySetLuaName(luaObj.LuaName, null))
+                    if (!luaObj.CanSetLuaName(luaObj.LuaName))
+                    {
                         luaObj.LuaName = string.Empty;
-                    editor.SelectedRoom.RemoveObject(editor.Level, obj);
+                        Editor.Instance.SendMessage("The value of Lua Name is already taken by another object", TombLib.Forms.PopupType.Error, true);
```
Nickelony: Use the available local `editor` variable rather than `Editor.Instance`

TombEditor/ObjectClipboardData.cs
```diff
@@ -78,9 +80,17 @@ public ObjectInstance MergeGetSingleObject(Editor editor)
...
+                    }
+                    else
+                    {
+                        luaObj.LuaName = luaObj.LuaName;
+                        editor.SelectedRoom.RemoveObject(editor.Level, obj);
+                    }
+
```
Nickelony: Remove newline

WadTool/WadActions.cs
```diff
@@ -1575,8 +1576,12 @@ public static WadMesh ImportMesh(WadToolClass tool, IWin32Window owner)
                     return null;
 
                 if (!dialog.FileName.CheckAndWarnIfNotANSI(owner))
+                {
+                    DarkMessageBox.Show(owner, "Filename or path is invalid. Please use standard characters.", "Wrong filename", MessageBoxIcon.Error);
                     return ImportMesh(tool, owner);
```
Nickelony: Remove newline

DarkUI/DarkUI/DarkUI.csproj
```diff
@@ -123,7 +123,4 @@
   <ItemGroup>
     <Content Include="Resources\grip_fill.png" />
   </ItemGroup>
-  <ItemGroup>
-    <PackageReference Include="System.Text.Encoding.CodePages" Version="5.0.0" />
```
Nickelony, line 127: Does the app run with this lib missing? It is used at the start of every `Program.cs` to allow the use of `Windows 1252` encoding across the apps.

Raildex: Where exactly? There is no usage of the CodePages Encoding

Nickelony: AFAIK if you build and run the app, it will crash immediately with an exception saying something like "Windows 1252 encoding is not available." But that might have been a .NET 5 error which may not be present in .NET 6, but please verify. If it doesn't crash, try and load a TRNG project in TombIDE and check if it can load and save any script files.

Raildex: Script files that are not UTF-8 have their non-UTF-8 characters replaced with `?` but no exception or anything.
Any other encoding than UTF should not be supported anyway.

Nickelony: The TRNG compiler (and the old TR4 one as well I think) can't compile UTF-8 documents, they can only be Windows 1252, otherwise the compiler throws an error.

Raildex: test it and require a PR change then.

Nickelony: Alright, so this was indeed a .NET 5 issue:
image

But in .NET 6, it seems to be fixed, sorry about that, I thought it was a general issue since moving to .NET Core, but they seem to have quietly fixed it.

Nickelony (approved):
Looks great! Thank you for sorting this out! :)

PR #1026, opened 2025-04-27
Smooth Terrain tool

Nickelony:


TombEditor/EditorActions.cs
```diff
@@ -465,6 +465,103 @@ public static void SmoothSector(Room room, int x, int z, SectorVerticalPart vert
...
+                }
+            }
+
+            SmartBuildGeometry(room, area);
+        }
+
+        private static void SmoothSectorCorner(Room room, int x, int z, SectorEdge edge, SectorVerticalPart vertical)
+        {
+            // Define the neighboring sectors and their corresponding edges based on the corner we're processing
+            (int X, int Z, SectorEdge Edge)[] neighbors = edge switch
```
Raildex: Consider turning this into a record struct.

Have you considered using a stack allocated Span for this?

TombEditor/Command.cs
```diff
@@ -1419,7 +1419,21 @@ static CommandHandler()
                 EditorActions.SmoothRandom(args.Editor.SelectedRoom, args.Editor.SelectedSectors.Area, -1, SectorVerticalPart.WS);
             });
 
-            AddCommand("SharpRandomFloorUp", "Sharp random floor up", CommandType.Geometry, delegate (CommandArgs args)
+			AddCommand("SmoothFloorTerrain", "Smooth floor terrain", CommandType.Geometry, delegate (CommandArgs args)
```
Raildex: The indendation does not match the surrounding code

TombEditor/EditorActions.cs
Nickelony: It's not a very re-usable tuple, so creating a struct for it seems rather wasteful. Also not sure if a span would improve anything here, as each collection is only 3 items.

Issue #1031, opened 2025-05-10
TE: Feature request - support data injection in TR1X and TR2X


lahm86:
Following on from a discussion with @Raildex, would it be feasible to have TE append additional injection data to the compiled level file for TR1X and TR2X? As a starting point, this would allow for having vertical rooms with indices greater than 254 (room above/below would each be a `short`, and `-1` would be used in place of `255` to mark "no room").

I'd be happy to work with you to generate the injection format. We have a (.NET Core 6) [library](https://github.com/LostArtefacts/TRXInjectionTool) we use for the OG injections, which I imagine could be plugged in to TE. Alternatively, the format is quite straight-forward so TE could have its own implementation of it. The latter may be preferable as the library has many other dependencies for OG injections, such as texture packing, animation data fixes etc, which would likely add unnecessary bloat in this context.

We'd need to adjust TR1X/TR2X to check the end of the level file and treat the remainder as injection data, but that should be straight-forward. The alternative would be having TE generate a separate injection file, but this would also need to be referenced in the game flow.

Happy to provide any further detail needed.

Lwmte:
Hello! Sure, it's been discussed on our discord server multiple times. Feel free to start a discussion about it there and/or create pull request, if you already have some prototype code for this feature.

Lwmte:
@lahm86 Hey, I believe this issue is obsolete now, since we now have dedicated TRX targets, or any extra work is still needed?

lahm86:
> [@lahm86](https://github.com/lahm86) Hey, I believe this issue is obsolete now, since we now have dedicated TRX targets, or any extra work is still needed?

Agreed! Happy to close this, future additions can be under their own issues for clarity. Thanks.

PR #1038, opened 2025-05-29
Add support for embedded TRX injections

lahm86:
Part of #1031.

This adds the ability to append TRX injection data to the end of generated level files for TR1 and TR2. The initial implementation adds support for having more than 255 rooms (in terms of vertical portals) by injecting sector overwrites where applicable.

Short demo here showing traversal of vertical portals at 254/255 and above.
https://youtu.be/_calOS6Kyyw

This wouldn't be available obviously yet to builders, but I'm not sure how you would like to proceed with adding TR1X/TR2X specifically, whether as separate game versions or just as a flag against TR1/TR2.

LMK your thoughts, happy to work with you to improve or revise this. Thanks!

### DIFF
--- a/TombLib/TombLib/LevelData/Compilers/Util/TrxInjector.cs
+++ b/TombLib/TombLib/LevelData/Compilers/Util/TrxInjector.cs
@@ -1,0 +1,167 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using TombLib.IO;
+using TombLib.Utils;
+
+namespace TombLib.LevelData.Compilers.Util;
+
+public static class TrxInjector
+{
+    private const uint _magic = 'T' | 'R' << 8 | 'X' << 16 | 'J' << 24;
+    private const uint _version = 2;
+    private const uint _injectionType = 0; // Implies no link to a TRX config option
+
+    public static void Serialize(TrxInjectionData data, BinaryWriterEx outWriter)
+    {
+        using var stream = new MemoryStream();
+        using var injWriter = new BinaryWriterEx(stream);
+
+        if (!WriteData(data, injWriter))
+        {
+            return;
+        }
+
+        var exportedData = stream.ToArray();
+        var zippedData = ZLib.CompressData(exportedData);
+
+        outWriter.Write(_magic);
+        outWriter.Write(_version);
+        outWriter.Write(_injectionType);
+
+        outWriter.Write(exportedData.Length);
+        outWriter.Write(zippedData.Length);
+        outWriter.Write(zippedData);
+    }
+
+    private static bool WriteData(TrxInjectionData data, BinaryWriterEx writer)
+    {
+        var chunks = new List<TrxChunk>()
+        {
+            CreateChunk(TrxChunkType.DataEdits, data, WriteEdits),
+        };
+
+        chunks.RemoveAll(c => c.BlockCount == 0);
+        if (chunks.Count == 0)
+        {
+            return false;
+        }
+
+        // Regular injections have applicability tests for OG levels. This is irrelevant
+        // for embedded injections.
+        writer.Write(0); // Number of tests
+        writer.Write(0); // Total length of tests
+
+        writer.Write(chunks.Count);
+        chunks.ForEach(c => c.Serialize(writer));
+        return true;
+    }
+
+    private static TrxChunk CreateChunk(TrxChunkType type,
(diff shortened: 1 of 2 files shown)


Lwmte:
Hi! Thank you for the pull request!
The way how you want file format extension to appear in level settings depends on your eventual goals with the extension functionality. If you plan just to inject extra data to extend original limits of the engine or original file format, it's better to implement it via checkbox option. If you eventually plan to extend editor's own functionality with TRX extra features (e.g. allow floor triangulation or colored lighting as in TR3 for TR1 and TR2), then probably better idea is to implement separate target.

lahm86:
Thanks, @Lwmte. I suppose going for separate targets would be best so that if we do wish to implement more editor-based changes in the futures it will be more seamless. Things like adding wind to rooms in TR1, and as you say potentially coloured lighting and suchlike. 
@Nickelony - sorry for the tag, but would you be able to help with this, as I believe you were looking at adding TR2X into TIDE? I guess we would need to decide if the current TR1 offering becomes TR1X exclusive (which IMO makes sense as it's already tagged as such in TIDE), and if there needs to be a separate non-TR1X option for anyone wanting to build for TombATI. I think it might be an idea to generate specific TR1X/TR2X catalogs too (perhaps that's a given with separate targets) as there are object types specific to TR1X, plus things like Lara's extended anim set in TR1, and raised limits for both TR1X and TR2X. I am happy to help with that.

lahm86:
Thanks, @Nickelony! No need to apologise, fully understand the need to prioritise 😌 
That sounds great then, LMK if you need anything from me. Thanks again.

Lwmte:
@lahm86 I must also mention that if we plan to support TRX-specific targets and level formats, then from your side it is needed to support these targets and level formats on a constant basis in TE code. TRX is updated fairly frequently, and Nickelony is already nearing his capacity to follow the changes in the scripting format.

TombLib/TombLib/LevelData/Compilers/Util/TrxInjector.cs
```diff
@@ -0,0 +1,174 @@
...
+        using BinaryWriterEx injWriter = new(stream);
+
+        if (!WriteData(data, injWriter))
+        {
+            return;
+        }
+
+        var exportedData = stream.ToArray();
+        using var outStream = new MemoryStream();
+        using var deflater = new DeflaterOutputStream(outStream);
```
Lwmte: We already have zlib integrated in TE codebase. Consider using zlib compression instead of zip.

lahm86: Fixed, thanks.

lahm86:
Hi @Lwmte. Just checking in to see if you need anything else from me before merging this one.

Nickelony:
Hi, sorry, I was also focusing on core features needed for 1.9 / 1.9.1 release. This and the fact that I'm also not feeling well this week. I'll try to review this PR on the weekend and resume work on adding TR2X support to TombIDE, but I'm afraid these features will have to wait for another release (e.g. 1.9.2).

rr-:
Hi @Lwmte, @Nickelony —

First, thank you again for the feedback so far, and for your continued work on TombEditor. I know everyone here is balancing open-source contributions alongside life and personal commitments, and I deeply respect that.

That said, I'm checking in again hoping we can finally move this PR forward. To recap briefly: we've followed your guidance from the previous Discord discussions and GitHub issue (#1031), specifically embedding injection data directly into the .phd/.tr2 files rather than external .bin files. We invested considerable time in this approach precisely because we believe it will significantly benefit builders and the broader community.

We're now approaching three months since the PR was submitted and even longer since the initial discussion. I fully appreciate that the recent 1.9.x releases required priority, and I genuinely understand that everyone, including @Nickelony, is doing their best under tight constraints. (I also caught the charity stream yesterday – big thanks to everyone for the kind words about TRX; they meant a lot!)

However, at this point, we're getting concerned that progress might stall indefinitely, especially since this PR is relatively small and purely additive. From our perspective, it's an initial foundation meant precisely to iterate from in future updates. Could you clarify if there's anything still blocking this from being merged, or if there are concrete adjustments needed from our end?

Naturally, we're committed to maintaining the TRX integration within TombEditor and fully owning that responsibility, but we also hope this PR won't set a precedent for extended waiting periods when it comes to smaller, incremental updates.

Thanks again – looking forward to hearing from you!

PR #1051, opened 2025-07-27
TombIDE: Add TR2X support

Nickelony:


TombLib/TombLib/LevelData/Enumerations.cs
```diff
@@ -14,8 +14,10 @@ public enum Game : long
             TR3 = 3,
             TR4 = 4,
             TR5 = 5,
-            TRNG = 16,
-            TombEngine = 18
+			TR1X = 11,
```
Lwmte: Since you have added new TR1X and TR2X enumerations, are you sure that TE and WT will function correctly now? Both TE and WT heavily rely on game versions to manage level building features (triangulated geometry, colored or grayscale lighting, and many many more). If builders will load TR2 wad into TR2X project, and so on, we need to make sure that it's resolved correctly.

TombLib/TombLib/LevelData/Enumerations.cs
```diff
@@ -25,17 +27,23 @@ public enum Game : long
...
         /// <summary> Wrapper for getting legacy game version, omitting both TRNG and TombEngine. Equal to deprecated TRVersion enum. </summary>
-        public static bool UsesMainSfx(this Game ver) => ver == Game.TR2  || ver == Game.TR3;
+        public static bool UsesMainSfx(this Game ver) => ver is Game.TR2 or Game.TR2X or Game.TR3;
 
         /// <summary> Base enumeration. Contains all possible game versions.
-        public static List<Game> AllVersions => Enum.GetValues(typeof(Game)).OfType<Game>().ToList();
+        public static List<Game> AllVersions => Enum.GetValues<Game>().ToList();
 
         /// <summary> Helper native (aka WadGameVersion) enumeration list. Can be used to populate various controls, like listbox. </summary>
-        public static List<Game> NativeVersions => AllVersions.Where(item => item != Game.TRNG).ToList();
```
Lwmte, line 34: I think that we are now missing something here.
`AllVersions.Where(item => item != Game.TRNG)` was here for a reason. TRNG is an extension of TR4, therefore, native version for TRNG is TR4. I guess the same principle should apply to TR1X and TR2X.

TombLib/TombLib/LevelData/Enumerations.cs
Nickelony: TR1X and TR2X entries are not currently used by Tomb Editor, only TombIDE uses it as a reference point. When the TombIDE project is `TR2X`, it creates a prj2 which targets `TR2` instead.

TombLib/TombLib/LevelData/Enumerations.cs
Nickelony: The collection approach is cleaner as it now also excludes TR1X and TR2X. The collection is essentially everything except `TR1X`, `TR2X` and `TRNG`.

Lwmte (approved):

PR #1064, opened 2025-09-20
Edit Batch Nodes

davidmarr:
- Added nodes to move moveables and statics
- Fixed descriptions
- Added full namespace for methods


Lwmte (approved):
Tested and approved, I have no objections.

PR #1082, opened 2025-11-22
Keypad nodes-display item

TrainWrack:
To be moved to animation editor lua code once it gets added. 

Revised Keypad nodes.

Setup is simpler and now only requires one node.

Uses Display Item Class. 

Adds  two new nodes. One to input, other to check the input.
image


TombLib/TombLib/Catalogs/TEN Node Catalogs/Keypad.lua
```diff
@@ -1,14 +1,30 @@
...
+[6] = 991,          -- TR5_Keypad_6
+[7] = 992,          -- TR5_Keypad_7
+[8] = 993,          -- TR5_Keypad_8
+[9] = 994,          -- TR5_Keypad_9
+["Failure"] = 995,  -- TR5_Keypad_Entry_No
+["Success"] = 996,  -- TR5_Keypad_Entry_Yes
+["Click"] = 644,    -- TR2_Click
+}
+
+local function keypadCreate(object, code)
```
Lwmte: Please use PascalCase for function names, regardless of local/global scope.

TombLib/TombLib/Catalogs/TEN Node Catalogs/Keypad.lua
```diff
@@ -1,14 +1,30 @@
...
--- !Section "User interface"
--- !Description "Creates a keypad."
--- !Arguments "NewLine, 80, Moveables, Keypad Object"
--- !Arguments "Numerical, 20, [ 1000 | 9999 ], Pass code"
--- !Arguments "NewLine, Volumes, Volume to use for the keypad"
-
-LevelFuncs.Engine.Node.KeypadCreate = function(object, code, volume)
+LevelVars.Engine.Keypad.ActivatedKeypad = nil
+
+local inventoryDelay = 0
```
Lwmte, line 4: Do we need to keep all these variables and a soundmap outside functions? Node script format was nice in a way it contained necessary code only within node functions. By exposing variables/tables in this way, we are breaking this format. Not saying it is wrong, just something I never planned to be done, because node functions were meant to be a wrapper for lua functions that can be directly used in scripts.

TrainWrack: I can move them in. I moved them out cause they would be created every frame but that’s ok.

PR #1091, opened 2025-12-13
TR4 Sequence Switch Node

TrainWrack:
Adds Sequence Switch Node

For sequence switch place the volume covering all 3 switches. name the switches any prefix such as SWITCH_1; SWITCH_2; SWITCH_3

Name doors DOOR_1; DOOR_2; DOOR_3; DOOR_4; DOOR_5; DOOR_6

Add the words SWITCH and DOOR to the node.

Refer attached project. 
[TR4 Objects.zip](https://github.com/user-attachments/files/26555530/TR4.Objects.zip)


TombLib/TombLib/Catalogs/TEN Node Catalogs/Objects.lua
```diff
@@ -0,0 +1,123 @@
...
+LevelVars.Engine.BurningFloor = {}
+LevelVars.Engine.BurningFloor.Active = false
+LevelVars.Engine.BurningFloor.Counter = BURN_DURATION
+LevelVars.Engine.BurningFloor.ItemName = nil
+LevelVars.Engine.BurningFloor.PlayerInVolume = false
+
+-- !Name "Burning Floor"
+-- !Section "Objects"
+-- !Description "Create a Burning Floor Object."
+-- !Arguments "NewLine, Moveables"
```
Stranger1992: No tooltips

TombLib/TombLib/Catalogs/TEN Node Catalogs/Objects.lua
```diff
@@ -0,0 +1,123 @@
...
+local BURN_DURATION = 1024
+
+local SOUNDS =
+{
+    FIRE_LOOP = 150,
+    SHATTER = 347 
+}
+
+local FIRE_OFFSETS =
+{   --displacement, rotation, size
```
Stranger1992: Are these offsets relative to the OG Burning Floor object?

TombLib/TombLib/Catalogs/TEN Node Catalogs/Objects.lua
```diff
@@ -0,0 +1,123 @@
+local START_DELAY = 0.25
+local BURN_DURATION = 1024
```
Stranger1992: I think the burning parameters should be included in the node itself. So the effect can be customised per burning floor.

TombLib/TombLib/Catalogs/TEN Node Catalogs/Objects.lua
```diff
@@ -0,0 +1,123 @@
...
+    end
+
+    TEN.Sound.PlaySound(SOUNDS.FIRE_LOOP)
+
+    if (math.abs(Lara:GetPosition().y - burningFloor:GetPosition().y) < 64) and normalizedTime >= START_DELAY and LevelVars.Engine.BurningFloor.PlayerInVolume then
+        Lara:SetEffect(Effects.EffectID.FIRE)
+        local ocb = burningFloor:GetOCB()
+        TEN.Sound.PlaySound(SOUNDS.SHATTER)
+        burningFloor:Shatter()
+        TEN.Flow.FlipMap(ocb)
```
Stranger1992: AS well as the OCB parameter (ie: choose flipmap number) and then set the OCB of the specified burning floor object to it .

Stranger1992 (dismissed):

TombLib/TombLib/Catalogs/TEN Node Catalogs/Objects.lua
TrainWrack: Yes.

TombLib/TombLib/Catalogs/TEN Node Catalogs/Objects.lua
TrainWrack: I can certainly make it an option.

Stranger1992: Also, watching the OG version back : the colour of the burning floor doesn't fade to black 

https://youtube.com/clip/UgkxVedmKiQO6JAusaCvA2GNosmuGOfgMkA3?si=5W7cAxM1Vt2-22-Z

Lwmte:
I believe this should be either implemented natively via C++, or by a Lua module operating on OnLoop callback, which we have now

TrainWrack:
> I believe this should be either implemented natively via C++, or by a Lua module operating on OnLoop callback, which we have now

We need to be able to insert code directly into object animations to effectively utilise onLoop callback. Otherwise we would still need a mechanism like nodes, module etc to attach the callback to the object.

TrainWrack:
I have removed all objects since I had another branch where they were added natively. Maybe we can leave this node as the setup is much easier.

PR #1114, opened 2026-01-30
Update TRX presets

lahm86:
This updates TR1X and TR2X presets to 1.1, superseding #1087 (unreleased). The TRX scripting folder is now embedded in the TRX exe, hence the removal of this as an "important folder" in TIDE (this was only added in #1087 anyway, so doesn't affect current release). Otherwise, the zip update is a straight-forward copy from the [current TRX stable release](https://github.com/LostArtefacts/TRX/releases/tag/trx-1.1).

Tested with legacy TR1X and TR2X projects, which correctly show as being outdated.
Tested creating new projects.

### DIFF
--- a/TombIDE/TombIDE.ProjectMaster/Forms/FormGameArchive.cs
+++ b/TombIDE/TombIDE.ProjectMaster/Forms/FormGameArchive.cs
@@ -82,7 +82,6 @@ namespace TombIDE.ProjectMaster.Forms
 				Path.Combine(engineDirectory, "cfg"),
 				Path.Combine(engineDirectory, "data"),
 				Path.Combine(engineDirectory, "shaders"),
-				Path.Combine(engineDirectory, "scripting"),
 			};
 
 			string[] importantFiles = new string[]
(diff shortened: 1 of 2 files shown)


TombIDE/TombIDE.ProjectMaster/Forms/FormGameArchive.cs
```diff
@@ -82,7 +82,6 @@ public void GenerateTR1Archive(string filePath, string readmeText)
 				Path.Combine(engineDirectory, "cfg"),
 				Path.Combine(engineDirectory, "data"),
 				Path.Combine(engineDirectory, "shaders"),
-				Path.Combine(engineDirectory, "scripting"),
```
Nickelony, line 85: So the `/scripting/` directory is not happening? Just want to verify this for the refactor I'm currently working on.

lahm86: This folder contained the internals for the Lua interface. It's still part of TRX, but it now gets embedded in the exe as it's not meant to be editable. Ref: https://github.com/LostArtefacts/TRX/pull/4567.

The recommended way for custom scripts - and indeed the way that's in place for OG levels - is to have a `data/scripts` folder. You can see the ones we currently ship for OG below. These are included in the updated presets for builder reference.
https://github.com/LostArtefacts/TRX/tree/develop/data/tr2/ship/data/scripts

PR #1120, opened 2026-02-07
Refactor Level Setup, Import and Rename forms + Refactor script generators

Nickelony:


Nickelony:
@copilot+claude-opus-4.6 Could you please review this PR and check for any regressions. Is this refactor production ready?

TombIDE/TombIDE.ScriptingStudio/LuaStudio.cs
```diff
@@ -106,117 +112,87 @@ private void IDEEvent_HandleSilentActions(IIDEEvent obj)
...
-					regex = new Regex(@"local\s+" + Regex.Escape(stringsVariableName) + @"\s*=");
-					var stringsStartLine = stringsEditor.Document.Lines.FirstOrDefault(line => regex.IsMatch(stringsEditor.Document.GetText(line)));
-
-					if (stringsStartLine == null)
-						return;
+				CreateGeneratedFiles(result.FilesToCreate);
+			}
+			catch
+			{
+				// Oh well...
```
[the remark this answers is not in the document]
Nickelony: Doesn't matter for now, we don't use logging in TombIDE. It is something to look into in the future though.

TombIDE/TombIDE.ProjectMaster/Forms/FormRenameLevel.cs
```diff
@@ -1,89 +1,66 @@
...
-				else // No errors
-					Height = 193;
-			}
+		private void AdjustFormHeight()
+		{
+			bool showNoErrors = !_initialState.ShowScriptError && !_initialState.ShowLanguageError;
+			bool showOneError = _initialState.ShowScriptError != _initialState.ShowLanguageError;
+
+			if (showNoErrors)
+				Height = 193;
```
Lwmte, line 59: Maybe it's worth dehardcoding these heights?

TombIDE/TombIDE.ProjectMaster/Forms/FormRenameLevel.cs
```diff
@@ -102,41 +79,28 @@ private void button_Apply_Click(object sender, EventArgs e)
...
+				var options = new LevelRenameOptions
+				{
+					NewName = textBox_NewName.Text,
+					RenameDirectory = checkBox_RenameDirectory.Checked,
+					RenameScriptEntry = checkBox_RenameScriptEntry.Checked
+				};
 
-				bool renameDirectory = checkBox_RenameDirectory.Checked;
-				bool renameScriptEntry = checkBox_RenameScriptEntry.Checked;
+				LevelRenameResult result = _levelRenameService.RenameLevel(
```
Lwmte, line 89: I believe it's not necessary to split this call into 4 lines.

TombIDE/TombIDE.Shared/SharedClasses/ScriptGenerator.cs
```diff
@@ -0,0 +1,172 @@
...
+			LoadCamera= 0, 0, 0, 0, 0, 0, 0
+			Horizon= {{horizonValue}}
+			""";
+	}
+
+	private static ScriptGenerationResult GenerateTombEngineScript(string levelName, string dataFileName, int ambientSoundId, bool horizon)
+	{
+		string horizonValue = horizon ? "true" : "false";
+
+		string gameFlowScript = $$"""
```
Lwmte, line 128: Not sure if it's worth it, but probably better idea is to offload script templates to external text files?

Nickelony: We would have to use a templating engine like Jinja2 or Scriban. It's not worth it for now.

TombIDE/TombIDE.ProjectMaster/Forms/FormRenameLevel.cs
Nickelony: I dont think we should care about this original WinForms code, as it will all be ported to WPF soon anyway. The point of this PR was to abstract the core code-behind logic away from the UI, so we can port it to WPF with minimal changes.

TombIDE/TombIDE.ProjectMaster/Forms/FormRenameLevel.cs
Nickelony: Yeah, you're right, but Forms code will be removed soon anyway and replaced with WPF ViewModels, so any changes to forms code right now are pointless, as it's dead code.
Code style in that case doesn't matter, unless it's an actual functional issue. It's all just temporary.

Lwmte (approved):

PR #1138, opened 2026-02-20
Refactored `RoomGeometry.cs`

Nickelony:


Nickelony:
@copilot Can you please review this PR to make sure the core logic of the RoomGeometry code hasn't changed due to the refactor, and that all the data it outputs is still identical to what it was before. Please point out any potential data output differences and whether they matter for the end result.

PR #1164, opened 2026-03-22
Fix numerous TE memory issues

Nickelony:
Addresses memory leaks and GC pressure from improper disposal patterns, missing event unsubscription, and redundant allocations in rendering paths.

### Resource Management

- **ObservableList<T>**: Remove unnecessary finalizer that was calling `Dispose()` for no reason
- **RenderingFont**: Split disposal logic to separate `ReleaseUnmanagedResources()` for proper destructor handling, add `GC.SuppressFinalize`
- **BinaryWriterFast**: Add `GC.SuppressFinalize` to skip redundant finalizer queue
- **ImportedGeometryTexture**: Implement `IDisposable` with `_disposed` guard to prevent GPU texture allocation after disposal in async `SynchronizationContext.Post` callback
- **ImportedGeometryManager/Panel3D**: Add null checks in `Dispose` to prevent `NullReferenceException` in design-time or early disposal

### Event Handler Cleanup

- **ImportedGeometryManager**: Store `ListChanged` handler reference for explicit unsubscription during disposal

### Rendering Optimizations

- **Panel3D**: Cache `_ghostBlockVertices` array and `_roomsDistanceCache` to eliminate per-frame allocations
- **WadTool panels**: Fix font disposal ownership - `RenderingFont.Dispose()` now owns `TextureAllocator`, remove redundant `_fontTexture?.Dispose()` calls

### Texture Hashing

Replaced `BinaryWriter` + `MemoryStream` + `ToArray()` with direct Blake3 hashing:

```csharp
using var hasher = Blake3Hasher.New();
Span<byte> header = stackalloc byte[8];
BinaryPrimitives.WriteInt32LittleEndian(header, Image.Size.X);
BinaryPrimitives.WriteInt32LittleEndian(header.Slice(4), Image.Size.Y);
hasher.Update(header);
hasher.Update(Image.ToByteArray()); // Returns underlying buffer, no copy
```


TombLib/TombLib/Wad/WadTexture.cs
```diff
@@ -20,14 +22,20 @@ public WadTexture(ImageC image)
...
-                Image.WriteToStreamRaw(ms);
-                Hash = Hash.FromByteArray(ms.ToArray());
-            }
+            // Hash image dimensions and pixel data directly without intermediate copies.
+            using var hasher = Blake3Hasher.New();
+            Span<byte> header = stackalloc byte[8];
+            BinaryPrimitives.WriteInt32LittleEndian(header, Image.Size.X);
+            BinaryPrimitives.WriteInt32LittleEndian(header.Slice(4), Image.Size.Y);
+            hasher.Update(header);
+            hasher.Update(Image.ToByteArray());
```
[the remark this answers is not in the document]
Nickelony: Incorrect, `.ToByteArray()` only returns the underlying `_data` field, rather than creating a copy of the data.

TombLib/TombLib/LevelData/RoomGeometry.cs
```diff
@@ -93,6 +93,10 @@ public void Build(Room room, bool useLegacyCode = false)
 
         private void ClearGeometryData()
         {
+            // Preserve previous capacity to avoid repeated internal array resizing.
+            int prevVertexCount = VertexPositions.Count;
+            int prevTriangleCount = TriangleTextureAreas.Count;
+
             VertexPositions.Clear();
             VertexEditorUVs.Clear();
             VertexColors.Clear();
```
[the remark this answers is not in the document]
Nickelony: Actually valid.

TombLib/TombLib/LevelData/ImportedGeometry.cs
```diff
@@ -44,11 +44,21 @@ private ImportedGeometryTexture(ImportedGeometryTexture other)
...
             DirectXTexture = other.DirectXTexture;
         }
 
+        public void Dispose()
+        {
+            DirectXTexture?.Dispose();
+            DirectXTexture = null;
+        }
+
         public override Texture Clone() => new ImportedGeometryTexture(this);
```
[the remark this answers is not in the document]
Nickelony: Fixed. `DirectXTexture` is not copied at all anymore.

TombLib/TombLib/LevelData/ImportedGeometry.cs
```diff
@@ -37,18 +37,27 @@ public ImportedGeometryTexture(string absolutePath)
...
             AbsolutePath = other.AbsolutePath;
             Image = other.Image;
             DirectXTexture = other.DirectXTexture;
         }
 
+        public void Dispose()
+        {
+            DirectXTexture?.Dispose();
+            DirectXTexture = null;
+        }
```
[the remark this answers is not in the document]
Nickelony: Fixed.

TombLib/TombLib/LevelData/ImportedGeometry.cs
```diff
@@ -31,22 +33,39 @@ public ImportedGeometryTexture(string absolutePath)
             if (SynchronizationContext.Current == null)
                 DirectXTexture = TextureLoad.Load(ImportedGeometry.Device, Image);
             else
-                SynchronizationContext.Current.Post(unused => // Synchronize DirectX, we can't 'send' because that may deadlock with the level settings reloader
-                DirectXTexture = TextureLoad.Load(ImportedGeometry.Device, Image), null);
+                SynchronizationContext.Current.Post(unused => {
+                    if (_disposed)
+                        return;
+
+                    DirectXTexture = TextureLoad.Load(ImportedGeometry.Device, Image);
```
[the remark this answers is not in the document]
Nickelony: Invalid.

TombLib/TombLib/LevelData/ImportedGeometry.cs
```diff
@@ -16,10 +16,12 @@
 
 namespace TombLib.LevelData
 {
-    public class ImportedGeometryTexture : Texture
+    public class ImportedGeometryTexture : Texture, IDisposable
     {
         public Texture2D DirectXTexture { get; private set; }
 
+        private bool _disposed;
+
```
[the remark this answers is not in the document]
Nickelony: Updated PR description.

TombLib/TombLib/Wad/WadTexture.cs
```diff
@@ -20,14 +22,20 @@ public WadTexture(ImageC image)
...
-            using (var ms = new MemoryStream())
-            {
-                var writer = new BinaryWriter(ms);
-                writer.Write(Image.Size.X);
-                writer.Write(Image.Size.Y);
-                Image.WriteToStreamRaw(ms);
-                Hash = Hash.FromByteArray(ms.ToArray());
-            }
+            // Hash image dimensions and pixel data directly without intermediate copies.
+            using var hasher = Blake3Hasher.New();
```
Lwmte, line 26: What's the advantage of `Blake3` hasher over `Hash.FromByteArray`?

TombLib/TombLib/LevelData/ImportedGeometry.cs
```diff
@@ -31,22 +33,39 @@ public ImportedGeometryTexture(string absolutePath)
             if (SynchronizationContext.Current == null)
                 DirectXTexture = TextureLoad.Load(ImportedGeometry.Device, Image);
             else
-                SynchronizationContext.Current.Post(unused => // Synchronize DirectX, we can't 'send' because that may deadlock with the level settings reloader
-                DirectXTexture = TextureLoad.Load(ImportedGeometry.Device, Image), null);
+                SynchronizationContext.Current.Post(unused => {
```
Lwmte, line 36: I don't fully understand what's happening here. This code prevents re-initialization of already disposed texture in race conditions?

DarkUI/DarkUI/Collections/ObservableList.cs
```diff
@@ -19,28 +19,13 @@ public class ObservableList<T> : List<T>, IDisposable
 
         #endregion
 
-        #region Destructor Region
-
-        ~ObservableList()
```
Lwmte, line 24: Please explain why `Dispose` methods were removed for this control. It's important because it's outside of TE codebase.

TombLib/TombLib/IO/BinaryWriterFast.cs
```diff
@@ -52,6 +52,8 @@ public void Dispose()
             _ptr = null;
             _startPtr = null;
             _baseStream = null;
+
+            GC.SuppressFinalize(this);
```
Lwmte, line 56: What's the purpose of this code?

TombLib/TombLib/LevelData/ImportedGeometry.cs
Nickelony: Yes, this is a very rare, but potential order of execution issue, as the posted lambda may happen after the object has been disposed. Extremely rare, but still good to add a defense mechanism just in case.

TombLib/TombLib/Wad/WadTexture.cs
Nickelony: Less heap allocation, no `MemoryStream` buffer, no `BinaryWriter` overhead, no final `ToArray()` copy, less GC pressure, slightly less CPU work. Both use `Blake3`, but this approach is more memory efficient with less overhead.

DarkUI/DarkUI/Collections/ObservableList.cs
Nickelony: The `Dispose()` methods weren't removed, they were restructured. The destructor was calling `Dispose()` for no reason, so the code has been simplified.

TombLib/TombLib/IO/BinaryWriterFast.cs
Nickelony: It avoids an unnecessary destructor call.
```cs
if (_handle.IsAllocated)
    _handle.Free();
```
Is already called in the `Dispose()` method, so the destructor does not have to be called again, saving one extra call. It also eases GC pressure ever so slightly, as you are marking it as "Hey, this is safe to be removed from memory now, no need to put it into the finalizer queue."

Nickelony:
@copilot+claude-opus-4.5 Could you please verify this PR is production ready and that there are no regressions because of the changes.

TombEditor/Controls/Panel3D/Panel3DDrawCollector.cs
```diff
@@ -144,12 +144,14 @@ Room[] CollectRoomsToDraw()
             // Collect rooms to draw
             var camPos = Camera.GetPosition();
             var roomsToDraw = CollectRoomsToDraw(_editor.SelectedRoom).ToArray();
-            var roomsToDrawDistanceSquared = new float[roomsToDraw.Length];
+
+            if (_roomsDistanceCache == null || _roomsDistanceCache.Length < roomsToDraw.Length)
```
Lwmte, line 148: If I understand correctly, even if now `_roomsDistanceCache` is not recreated every draw call, it can grow infinitely, depending on the overall room count, and is never cleared. For example, if I load a level with 700 rooms and turn on "Draw all rooms" mode, `_roomsDistanceCache` will now grow to 700 units and will stay like that, even if I will close the level and make a new one with just 1 room.

TombLib/TombLib/LevelData/ImportedGeometry.cs
```diff
@@ -31,22 +33,39 @@ public ImportedGeometryTexture(string absolutePath)
...
+                    if (_disposed)
+                        return;
+
+                    DirectXTexture = TextureLoad.Load(ImportedGeometry.Device, Image);
+                }, null);
         }
 
         private ImportedGeometryTexture(ImportedGeometryTexture other)
         {
-            DirectXTexture = other.DirectXTexture;
```
Lwmte, line 40: Is this assignment now happening elsewhere, or not needed?

TombEditor/Controls/Panel3D/Panel3DDrawCollector.cs
Nickelony: You're right, I've pushed a fix for this.

TombLib/TombLib/LevelData/ImportedGeometry.cs
Nickelony: This is an ownership problem, since `DirectXTexture` **_needs_** to be disposed, you can't make 2 instances of `ImportedGeometryTexture` share the same reference. I remember checking this and found no issues with this being removed, but I will check again later.

Nickelony: @copilot+claude-opus-4.6 From my initial analysis, the clones of the textures are purely for data driven operations, and the DX texture itself doesn't need to be copied, but I please double-check this claim and whether the DX texture not being copied causes any new issues/regressions.

MontyTRC89 (approved):

PR #1181, opened 2026-04-03
Adeed support for .prj and .wad files in FileAssociation

Checkm8Croft:


Lwmte:
Hi! Thank you for the pull request! @Nickelony please take a look at this PR when you have time, thanks!

FileAssociation/FormMain.cs
```diff
@@ -17,7 +19,9 @@ public FormMain()
 
 		protected override void OnLoad(EventArgs e)
 		{
+         _wasPRJChecked = checkBox_prj.Checked = Association.IsAssociatedWith(".prj", DefaultPaths.TombEditorExecutable);
```
Nickelony: Indentation inconsistency, we use tabs rather than spaces in this part of the code.

FileAssociation/FormMain.cs
```diff
@@ -54,8 +70,10 @@ private void button_Close_Click(object sender, EventArgs e)
 			=> Close();
 
 		private void checkBox_CheckedChanged(object sender, EventArgs e)
-			=> button_Apply.Enabled = checkBox_prj2.Checked != _wasPRJ2Checked
-				|| checkBox_wad2.Checked != _wasWAD2Checked
-				|| checkBox_trproj.Checked != _wasTRPROJChecked;
+          => button_Apply.Enabled = checkBox_prj.Checked != _wasPRJChecked
```
Nickelony: Another indentation inconsistency. Use tabs instead of spaces. Additionally, the OR chain below should have 4 tabs instead of 5

FileAssociation/Program.cs
```diff
@@ -38,20 +38,30 @@ private static void OpenGUI()
 
 		private static void RunAssociation()
 		{
+            AssociatePRJ();
```
Nickelony: Again, use tabs instead of spaces please.

FileAssociation/Program.cs
```diff
@@ -38,20 +38,30 @@ private static void OpenGUI()
...
 			AssociatePRJ2();
+			AssociateWAD();
 			AssociateWAD2();
 			AssociateTRPROJ();
 		}
 
-		private static void RunAssociationWithBinaryArgs(string args)
+       private static void RunAssociationWithBinaryArgs(string args)
```
Nickelony: Same as above.

FileAssociation/Program.cs
```diff
@@ -38,20 +38,30 @@ private static void OpenGUI()
...
 		{
-			bool associatePRJ2 = int.Parse(args[1].ToString()) != 0;
-			bool associateWAD2 = int.Parse(args[2].ToString()) != 0;
-			bool associateTRPROJ = int.Parse(args[3].ToString()) != 0;
+			bool associatePRJ = int.Parse(args[1].ToString()) != 0;
```
Nickelony: ~~The order of args has changed here, we can't do this due to compatibility reasons with the old installer. Please restore the previous PRJ2, WAD2 and TRPROJ order and add the new PRJ and WAD entries at the end.~~

FileAssociation/Program.cs
```diff
@@ -61,7 +71,9 @@ private static void RunAssociationWithBinaryArgs(string args)
 
 		private static void RunDeassociation() // Is this even a real word?
 		{
+         Association.ClearAssociations(".prj");
```
Nickelony: Indentation inconsistency.

Nickelony (changes requested):

Nickelony:
I would not merge this PR for 2.0 release, as we need to design new icons for legacy PRJ and WAD files. We don't want them to share the same icons as the modern ones. Once all my comments are addressed, I will merge this PR for a 2.1 release and add new icons for this myself.
Additionally, please re-target to our `develop` branch rather than `master`. Thank you! :)

FileAssociation/Program.cs
Nickelony: In fact, this is a deeper problem we need to look into with the old installer. You are okay to ignore this for now.

Checkm8Croft:
Sorry if the Italian Translation of TombIDE is here, GH put it there

Nickelony:
> Sorry if the Italian Translation of TombIDE is here, GH put it there

Please don't translate TombIDE, as that translation API was only experimental and it was never meant to go public. It will be removed in a future update. There's no way to change the language right now anyway.

PR #1216, opened 2026-05-01
Heat haze and DOF controls

Lwmte:
Introduces new "Distortion" blend mode for TEN target and flyby DOF controls and preview.
Complementary PR for https://github.com/TombEngine/TombEngine/pull/1983 .

### DIFF
--- a/TombEditor/Controls/FlybyTimeline/Preview/FlybyFrameState.cs
+++ b/TombEditor/Controls/FlybyTimeline/Preview/FlybyFrameState.cs
@@ -2,6 +2,7 @@
 
 using System.Numerics;
 using TombLib;
+using TombLib.LevelData;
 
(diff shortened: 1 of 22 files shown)


TrainWrack:
Tested. The DoF preview in editor is great and it helps explain how it actually works too. The effect can be tuned in realtime in the option box.

I could not test distortion on room geometry as the geometry would not draw. Same with objects, the faces do not draw when distortion blend mode is set in wadtool. No mesh is visible either in wireframe mode. Its probably similar to the issue I was having with the particles. I deleted and rebuild shaders.

TrainWrack (approved):
Tested the DoF preview in editor and ingame with consistent results.
Distortion effect acts as a mask. Tested on room geometry and moveable object.

PR #1217, opened 2026-05-02
Fix DarkUI's floating toolbox on higher DPI monitors

Stranger1992:
This pull request addresses high-DPI scaling issues by modifying how DPI awareness is handled in the application and by explicitly setting scaling properties for a custom control. These changes aim to provide more predictable UI rendering, particularly on high-DPI displays.

**DPI scaling and rendering adjustments:**

* Changed the application's DPI mode to `DpiUnaware` in `Program.cs` to prevent automatic DPI scaling, ensuring consistent rendering across different display settings.
* Set `AutoScaleMode` to `None` and defined `AutoScaleDimensions` in the `DarkFloatingToolBox` control to disable automatic scaling and specify a fixed design-time DPI, helping maintain consistent control sizing.


Nickelony (changes requested):
I cannot approve switching to `DpiUnaware`, as this is a major breaking change that is most definitely going to break everything. There needs to be another way to solve this issue, right?

Lwmte:
@Nickelony I believe that on contrary, `HighDpiMode.SystemAware` setting may be responsible for many UI bugs that are experienced by people with DPI set above 100%, such as misaligned controls. We need to test it and if it turns out to be true, we also need to use `HighDpiMode.DpiUnaware` for all apps in TE suite until we fully migrate to WPF.

Lwmte (approved):

PR #1234, opened 2026-05-29
Consolidate textures

Lwmte:
This PR allows to consolidate disjointed texture pieces for a given object into a single set of texture pages.


TrainWrack:
Did a quick test on Consolidate texture branch. Its good but not quite what wadmerger had achieved. It consolidates textures only on one object. Which is already an improvement but for Lara which spans multiple objects, it would be good, if we can select multiple objects and then consolidate into one. 

Also, maybe we can auto consolidate, when copying objects out of TR files. Once again, if I consolidate the textures and there are multiple objects which use the same textures, they each get their own copy of the said texture.

TrainWrack (approved):
Merge conflicts to be resolved.

But the branch works as described.

Tested for TR levels and wads both.

PR #1238, opened 2026-06-03
Multiple slot remaps

Lwmte:
Allows to specify multiple TEN remap entries in the trcatalog. To do it, format xml entry like this:

`<moveable id="86" name="Heavy swinging thing / Drill bit / Swinging brazier" essential="false" ten="SLOT1, SLOT2, SLOT3" />`


Stranger1992 (approved):
Works as intended. Now: multiple objects from the classics that are shoved into the same slot can be remapped easier to Tomb Engine:

Image

* Sounds still correctly remap to the Tomb Engine sound catalog.
* Animation data (state numbers , next frame , next anim ect) correctly carry-across.

Approved. Thankyou @Lwmte

PR #1248, opened 2026-06-15
Tomo traps wave2 properties

Nemoel-Tomo:
added property catalogue to Falling Spikes, Fan, Large fan, and disk shooter


TrainWrack:
@Lwmte We should think of a naming standard for the Property xmls.

I am not a fan of internal naming. 

Maybe we can do something like NumericalID_Name
like 321_DISK_SHOOTER.xml

Stranger1992:
> @Lwmte We should think of a naming standard for the Property xmls.
> 
> I am not a fan of internal naming. 
> 
> Maybe we can do something like NumericalID_Name
> like 321_DISK_SHOOTER.xml

I disagree. Having a naming convention like `ID_DOG` without the slot number is better because one day , the numbered slot system won't be used iirc

Lwmte:
I propose using object name as well, but without `ID_` prefix.

Nemoel-Tomo:
Cataloge names fixed without ID_

TombLib/TombLib/Catalogs/TEN Property Catalogs/FIRE_PENDULUM.xml
```diff
@@ -0,0 +1,28 @@
+<propertyCatalog>    
+	<moveable id="DISK_SHOOTER">
```
Lwmte: I believe this is a copypaste mistake. DISK_SHOOTER.

TrainWrack (commented):
- [x] The sounds for Overhead_pulley do not map correctly. Should be sound 670 in TEN ALL SOUNDS.
- [x] The sounds for TR1 Slamming door weremissing in the repo. Mapped out to use TR2 sound slots. 

All other prorperties work ok.

PR #1256, opened 2026-07-19
Update TRX presets

lahm86:
This updates the TR1X and TR2X presets to 1.9.3. It also resolves being unable to use sound IDs in TRX levels above the OG sound map size without also having to define all other lower IDs.

The presets are now slimmer in terms of not including all the OG level injections, and a stripped back game-flow.


Nickelony (approved):

Nickelony:
Thank you! I will make sure to add automatic level config generation in either 2.1 or 2.2. That should help with creating gameflow level instances within an empty gameflow file. :)

PR #1267, opened 2026-07-26
Always relight all rooms when entering light preview mode

Nickelony:


Nickelony:
@copilot+gpt-5.6-terra Could you please review this PR and make sure there are no regressions, and that the implementation is correct?

Nickelony:
@copilot+gpt-5.6-terra No, this isn't right, it can take multiple seconds to relight all rooms in the entire level, making the UI hang while doing so, and we can't do that.

Nickelony:
That wasn't the fix. Closing this PR.

PR #1289, opened 2026-09-09
Fix TR1 and TR2 pistol ammo

lahm86:
This fixes pistol ammo in TR1 and TR2 not being converted to the expected engine type when compiling.


Nickelony (approved):
Looks correct!

PR #1292, opened 2026-09-17
Fix TRX portals

lahm86:
This fixes portal overwriting in TRX levels where room numbers exceed 255. The previous implementation didn't account for solid floors or ceilings that sit within a larger portal.

This fix has been tested by a builder, conversation in Discord.


Nickelony (approved):
LGTM!

